PR #7, opened 2015-05-16
track applied patches and only remove/redownload/re-patch package if patchset changed

yched:
This writes ['extra']['patches_applied'] for each patched package in vendor/composer/installed.json, and uses the information to determine whether a package should be removed/redownloaded/repatched, to avoid doing that on every "composer install"


cweagans:
I'll look at this in more detail later, but my impression is generally positive at this point.

yched:
Cool :-)

rebased after the other PRs were merged

cweagans:
Been thinking about this and I'm happy with it. Thanks!

PR #9, opened 2015-05-26
check patches also on composer update

yched:
checkPatches() is there so that a package gets re-doawnloaded and repatched if its patch set changes, even if we keep the same version of the module.

This case applies also on composer update - we want to reapply patches if they changed, iven if the "update" doesn't bring a new version of the package.


cweagans:
:+1:

PR #10, opened 2015-07-13
Documentation update: post-package-install is no longer necessary

greg-1-anderson:
In my testing, and as shown in the rest of the README, the post-package-install directive does not appear to be required inside the composer.json file with this patch manager any longer.


cweagans:
:+1:

PR #14, opened 2015-08-24
Add optional git support

LionsAd:
This adds very basic and optional git support to composer-patches.

The use case is that we want the history of the patches be visible in the GIT history.

Also having each package committed separately is a nice side-effect.

To use:

```
export COMPOSER_PATCHES_USE_GIT=1
export COMPOSER_PATCHES_GIT_COMMIT_MESSAGE_PREFIX="Ticket #12345: "
composer install
```

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -43,6 +43,10 @@ class Patches implements PluginInterface, EventSubscriberInterface {
    * @var array $patches
    */
   protected $patches;
+  /**
+   * @var bool $useGit
+   */
+  protected $useGit = FALSE;
 
   /**
    * Apply plugin modifications to composer
@@ -55,6 +59,7 @@ class Patches implements PluginInterface, EventSubscriberInterface {
     $this->io = $io;
     $this->executor = new ProcessExecutor($this->io);
     $this->patches = array();
+    $this->useGit = getenv('COMPOSER_PATCHES_USE_GIT') == '1';
   }
 
   /**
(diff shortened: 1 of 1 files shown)


cweagans:
I'm not sure how well this will work with https://github.com/cweagans/composer-patches/blob/master/src/Patches.php#L113. When you update a patch set for a given package, that package is removed from disk and re-downloaded.

cweagans:
More generally, I'm not sure why you'd want to do this at all Given that your patch list is committed to composer.json/composer.lock, the history is relatively visible by looking at the history for that one file. Moreover, dependencies downloaded by Composer shouldn't be committed to the top level Git repository (and each of the dependencies should have their own VCS repo). Why not just push a tag to the dependency repositories, and then change what tag you're pulling down in your composer.json?

I may be missing something because I'm very confused by this.

LionsAd:
@cweagans 

It works fine with packages that are removed.

For various reasons we commit the sites/all/modules/contrib to GIT and use composer solely to update it.

We are using https://github.com/douggreen/drupal-composer-installer for that.

If this approach does not fly:

- What about a PatchEvent that is triggered before and after applying each patch?

Would that be acceptable?

Then we can add it to drupal-composer-installer itself.

cweagans:
I'd be more in favor of a PatchEvent, yes. That's much more useful to a wider group of people, I think.

LionsAd:
Okay, so I would say we send in the PatchEvent:

```
       try {
+        $this->eventDispatcher->dispatch(NULL, new PatchEvent(PatchEvents::PRE_PATCH_APPLY, array(), array(), $package, $install_path, $url,
         $this->getAndApplyPatch($downloader, $install_path, $url);
+        $this->eventDispatcher->dispatch(NULL, new PatchEvent(PatchEvents::POST_PATCH_APPLY, array(), array(), $package, $install_path, $url
         $extra['patches_applied'][$description] = $url;
       }
```

Does that sound good?

cweagans:
Works for me. Note that you can get the install path from `$package` if you send the full package object, so that may not be strictly necessary, but I'm not opposed to including it if it makes your life easier.

cweagans:
Also note that since this and the PHP 5.3 compatibility PR have gone in since the last tag, it's probably about time to do a new release, so you'll have a tag you can point at if you want.

LionsAd:
Oh, right. Yes lets keep that clean.

PR #15, opened 2015-08-24
Add event for applying patches

LionsAd:


cweagans:
:+1: This looks really good.

Thanks for setting up phpunit, too! Been meaning to get around to that. The rest of the plugin has no tests! :(

Issue #19, opened 2015-10-27
Add checksum to patch definition


badjava:
In addition to the patch title and URL, there should be an optional checksum to verify the file. This is critical for security particularly when applying patches from d.o or other external sources.

Drupal make [supported this](http://www.drush.org/en/master/make) with md5.

cweagans:
If you're concerned about it, is there any reason you can't just download the patch and keep it locally? That way, you're not depending on external sources and verification isn't necessary.

If this is really 100% necessary, we'll have to find some way to do it that satisfies these requirements:

* No BC break for existing users
* Completely optional
* Doesn't add a massive amount of complexity to the plugin

For the record: I'm not opposed to this, but there are definitely some other things ahead of this on my priority list.

badjava:
Thank you for your quick response and I fully understand there are other priorities ahead of this. I think it will be required, especially for enterprise usage. I noticed some potential BC issues with the patch definition format so I wanted to make sure it was on the radar.

cweagans:
> especially for enterprise usage.

That's actually the reason I ask. We're using it internally @ NBCUniversal, but when we did the work on this plugin, we didn't discuss that particular requirement.

@ericduran: If implemented, is this something we would use?

skwashd:
@cweagans I wrote the patch that added checksum support to drush make.  Being able to reference online patches and be confident that they haven't changed is a key requirement for several of the clients I've worked with over the years.

The checksums in drush make have always been optional and were implemented in a non BC breaking way.  I am yet to dig into the code, but I think it should be possible to do the same here.

IMO checksums should be used for any patch, even from "trusted" sources.  It ensures that nothing has changed since it was added to the composer.json.  Maybe it is a small tweak to a gist or it could be something more malicious.  Either way users should have the option of this change breaking a build and a human investigating it.

Given the current state of cryptographic standards I propose that we use SHA-256 checksums.

cweagans:
@skwashd Like I said, I'm not opposed to adding it, but it's not a high priority for me right now. If a PR is opened that satisfied the criteria in https://github.com/cweagans/composer-patches/issues/19#issuecomment-151601734, I'll happily review and merge.

cweagans:
(Also, yes - not md5 please. SHA-1 or SHA-256 are both fine. SHA-1 is okay in this case since it's more of a consistency check than a security feature - see https://en.wikipedia.org/wiki/SHA-1#Data_integrity for related discussion from Linus)

[...]

PR #22, opened 2015-11-06
Add support for local patches.

tobiasbaehr:
I tried it with file:///patches/mypatch.patch but this do not work with parse_url.

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -276,12 +276,19 @@ class Patches implements PluginInterface, EventSubscriberInterface {
    * @throws \Exception
    */
   protected function getAndApplyPatch(RemoteFilesystem $downloader, $install_path, $patch_url) {
-    // Generate random (but not cryptographically so) filename.
-    $filename = uniqid("/tmp/") . ".patch";
 
-    // Download file from remote filesystem to this location.
-    $hostname = parse_url($patch_url, PHP_URL_HOST);
-    $downloader->copy($hostname, $patch_url, $filename, FALSE);
+    // Local patch file.
+    if (file_exists($patch_url)) {
+      $filename = $patch_url;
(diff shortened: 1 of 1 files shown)


cweagans:
:+1:

yched:
Awesome, support for local patches would be definitely precious.

However, we'd need a way to trigger a module repatch if the content of the local patch file changes (which is not really an issue with remote patches, since we can assume "same url = same content")

That probably means storing a hash of the file content in installed.json when applying the patch,, so that we can compare the next time around.

yched:
Oops, crosspost :-)

cweagans:
@yched You're right. This is a good first pass, though. We can implement the re-patch functionality in another PR. I think it's reasonable to get a hash of every patch (local or remote) and store it. That way, we know if a remote hash has changed too. Getting that into place simplifies https://github.com/cweagans/composer-patches/issues/19 as well.

That said, I want to refactor the plugin quite a bit. `Patches.php` has like 10 separate jobs that it's doing, and if it were broken out into smaller pieces, it'd be a lot more testable, IMO. The bigger this plugin gets, the more scared I am to just merge a PR. I don't want to go too much further down this path without having a pretty solid test suite in place.

yched:
Also, I cant test right now, but according to http://php.net/manual/en/function.file-exists.php,  file_exists() also works with 'http://...' urls ? So the code here doesnt actually differentiate between local and remote patches ?

cweagans:
Ugh. wtf php.

How about `is_file() && is_readable()` ?

yched:
Sorty, my bad.
http://php.net/manual/en/wrappers.http.php says "Supports stat() : No". 

So file_exists() should be ok :-)

PR #24, opened 2015-11-28
Adds optional support for SHA1 checksums for patches

skwashd:
**Breaks backwards compatibility** - basis for new 2.x branch
Changes json format to accommodate the SHA1 checksum
Clean up and fix output handling for shell commands.


skwashd:
@cweagans I found that it wasn't possible to satisfy all of the requirements in #19.  I'm proposing this as the basis of 2.x as the json schema has changed.

yched:
Nice !

I think supporting an extended definition for each patch (array of key/values instead of just name -> url) would be nice, as it allows more metadata (explanations, issue number...) to help maintaining the patch set.

It should be possible to support both syntaxes and not break BC, though (with is_string() / is_array() checks) ?

cweagans:
Honestly, I see 2.x being more of a rewrite than anything. Right now, this plugin has like ~2% test coverage or something like that, and the main plugin class is doing way too much. As that class expands, I'm getting less comfortable with adding new things because every subsequent change has an increasing set of potential breakage.

I agree with @yched - should be able to do a check for an array and then check for a sha1. Alternatively, you could do something like

```
"My patch description": "2346ad27d7568ba9896f1b7da6b5991251debdf2$$http://example.com/mypatch.patch"
```

or similar and `explode()` on whatever separator is appropriate, though that feels pretty hacky.

cweagans:
#447 includes sha256 hashes for all patches (calculated on first use if they aren't provided)

Issue #26, opened 2015-12-17
Package deleted but not reinstalled and patched


j-hall-ctidigital:
This is happening on both Composer install\update after I've initially updated the patches file. The next time I run an install or update however the package is reinstalled and patched as expected. 

This is the terminal output when the package is not reinstalled:

![Image of terminal output](http://puu.sh/lZ5h8/09e6c647fe.png)

The extra section of my composer.json

``` json
"extra": {
        "installer-paths": {
            "web/core": ["type:drupal-core"],
            "web/modules/contrib/{$name}": ["type:drupal-module"],
            "web/profiles/contrib/{$name}": ["type:drupal-profile"],
            "web/themes/contrib/{$name}": ["type:drupal-theme"],
            "drush/contrib/{$name}": ["type:drupal-drush"]
        },
        "merge-plugin": {
            "include": [
                "web/modules/custom/*/composer.json"
            ],
            "recurse": true,
            "replace": false,
            "merge-extra": false
        },
        "patches-file": "composer.patches.json"
    }
```

and my composer.patches.json

``` json
{
  "patches": {
    "drupal/devel": {
      "Patch title": "https://www.drupal.org/files/issues/wrong_dependencies_are-2629828-6.patch"
    }
  }
}
```

cweagans:
Sorry for the slow response here. I've been insanely busy for the past month :(

First, please ensure that you're running the latest Compose, as that may affect this functionality.

If the problem persists, are you able to provide a complete `composer.json` that reproduces this problem reliably? If so, I can dig into it and fix it.

grasmash:
I'm also running into this problem. I can reproduce it in the following way:

* Add or remove a patch for drupal/drupal
* Run `composer update`. 
* Docroot is deleted but not recreated
* Run `composer update` again and it's all back!

Example file:

```
{
  "name": "example",
  "description": "Example config for a D7 site using composer based build process.",
  "repositories": [{
    "type": "composer",
    "url": "https://packagist.drupal-composer.org"
  }],
  "require": {
    "composer/installers": "~1.0",
    "cweagans/composer-patches": "~1.0",
    "derhasi/composer-preserve-paths": "0.1.*",
    "drupal/drupal": "7.*",
    "drupal/views": "~7.0",
    "drupal/ctools": "~7.0"
  },
  "conflict": {
    "drupal/core": "8.*"
  },
  "scripts": {
    "post-create-project-cmd": [
[...]

PR #30, opened 2016-02-17
Patches defined in dependencies are not applied unless the dependency is updated.

mathieuhelie:
I've been struggling to rebuild a project of ours that defines patches for drupal modules in a custom drupal profile. The patch gathering process was only reading patches from the root package and packages that receive an update or install. The modified plugin in this PR stores patches definition in installed.json in $this->installedPatches, then adds them to the patch list unless the package that installed them is itself updated.

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -60,6 +60,7 @@ class Patches implements PluginInterface, EventSubscriberInterface {
     $this->eventDispatcher = $composer->getEventDispatcher();
     $this->executor = new ProcessExecutor($this->io);
     $this->patches = array();
+    $this->installedPatches = array();
   }
 
   /**
@@ -95,6 +96,9 @@ class Patches implements PluginInterface, EventSubscriberInterface {
 
       foreach ($packages as $package) {
         $extra = $package->getExtra();
+        if (isset($extra['patches'])) {
+          $this->installedPatches[$package->getName()] = $extra['patches'];
+        }
         $patches = isset($extra['patches']) ? $extra['patches'] : array();
         $tmp_patches = array_merge_recursive($tmp_patches, $patches);
       }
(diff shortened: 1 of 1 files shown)


mathieuhelie:
The way the plugin currently decides whether or not to Uninstall and repatch a package also means that new patches defined by dependencies won't trigger an Uninstall since the updated composer.json of the dependencies hasn't yet been read in the PRE_INSTALL_CMD and PRE_UPDATE_CMD events.

Ideally the uninstalling of patched packages should take place after the dependencies have been resolved, and before the PRE_PACKAGE_INSTALL and PRE_PACKAGE_UPDATE events. I tried the InstallerEvents::POST_DEPENDENCIES_SOLVING event as the trigger but it failed to run, it would probably require substantial refactoring.

cweagans:
Sorry for the slow response on this. I'm not 100% sure I understand the problem you're running into. Could you please provide a `composer.json` that causes the problem, along with any dependencies' `composer.json` and walk me through what you're seeing? If it's easier, I'm happy to jump on a Join.me with you or something.

mathieuhelie:
So our composer.json triggering the issue are pretty complicated, but here's an example scenario:

Root package
- requires ourvendor/ourprofile
- requires drupal/views ~7.3.1
- adds patch A for drupal/views

Package ourvendor/ourprofile
- requires drupal/views ~7.3.1
- adds patch B for drupal/views

If running a composer update on our root package, composer may detect a version increase in drupal/views and trigger an update command on the package. The patches plugin will then gather patches from the root composer.json but will skip the patches from ourvendor/ourprofile because as far as composer tells it there has been no change in that package and it is irrelevant.

Since composer stores the composer.json for ourvendor/ourprofile in installed.json, we can gather all the patches for drupal/views from that source.

chrfritsch:
This is exactly what we urgently need for our drupal thunder distribution

grasmash:
This is the cause of #26.

cweagans:
:+1:

grasmash:
I have tested that this works on initial install, update, and with successive updates. It appears to work with root patches as well as patches that belong to dependencies.

anotherjames:
I honestly don't know why, but I still have the same problem. I'm using commit ee6a6d0 of this plugin. I had 2 patches listed for a required package (config_devel drupal module, as it happens, if that makes any difference). I added a 3rd patch, and ran composer update. The package was removed, and not reinstalled. I ran composer update a 2nd time, this time it was reinstalled and patched with all 3 patches.

PR #31, opened 2016-02-23
Throw a new exception after catching Cannot apply patch , so that Com…

pivulic:
### Problem
When a patch fails, an exception is thrown from: https://github.com/cweagans/composer-patches/blob/master/src/Patches.php#L349

The exception is then catched in https://github.com/cweagans/composer-patches/blob/master/src/Patches.php#L264

During the Composer install, the patch fails to be applied and the user is informed about the error, but the install process continues.

### Expected Result
If a patch fails to be applied, an exception should be thrown so that Composer fails the install and the user is notified about the error. Today, if the Composer installs several packages, the user might not wait for the entire install to finish and might therefor miss that a patch failed.

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -263,7 +263,9 @@ class Patches implements PluginInterface, EventSubscriberInterface {
       }
       catch (\Exception $e) {
         $this->io->write('   <error>Could not apply patch! Skipping.</error>');
-        throw new \Exception("Cannot apply patch $url");
+        if(getenv('COMPOSER_EXIT_ON_PATCH_FAILURE')) {
+          throw new \Exception("Cannot apply patch $url");
+        }
       }
     }
     $localPackage->setExtra($extra);
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -263,6 +263,7 @@ class Patches implements PluginInterface, EventSubscriberInterface {
       }
       catch (\Exception $e) {
         $this->io->write('   <error>Could not apply patch! Skipping.</error>');
+        throw new \Exception("Cannot apply patch $url");
       }
     }
     $localPackage->setExtra($extra);


cweagans:
I agree with the intent of this patch, but I think it needs a bit of work first. Some people may just want to continue, some may want it to stop. For now, let's leave the default behavior as-is, but add a way for users to opt-in to throwing another exception so that the `composer install/update` stops.

I'm thinking something like this:

```
  "extra": {
    "patches": {
      "vendor/package": {
        "My patch": "http://example.com/mychanges.patch"
      },
      "_config": {
        "exit-on-patch-failure": true,
      }
    }
  }
```

@pivulic Do you want to do this?

moleman:
One suggestion would be to be able to set "exit-on-patch-failure" as an environment variable. Then you can fail or not fail "composer install / update" depending on the environment. In a build environment it's important to fail the build if a patch is not correctly applied.

Example:
```
export CP_EXIT_ON_PATCH_FAILURE=1
composer install
```

pivulic:
:+1:

cweagans:
@moleman, @pivulic :+1: from me. That's even better. The main point is that it should be opt-in. If you're going the environment variable route, please prefix with `COMPOSER_` to be consistent with other Composer environment vars (so maybe `COMPOSER_EXIT_ON_PATCH_FAILURE`).

pivulic:
What about now? :)

cweagans:
Committed in https://github.com/cweagans/composer-patches/commit/0d75c63b6de66144517a1da9084422b78c2ff251 -- added a space, changed some copy, and added some detail to the commit message, but it's all still credited to you. Thanks!

pivulic:
Thanks! By the way, when will this be included in a new release? :)

cweagans:
When there's enough to justify a new release - 1-2 months tops.

pivulic:
Would it be possible to release a minor version at least, i.e. 1.3.1? :) That way, we can use your package from https://packagist.org/ instead of our fork in our projects.

cweagans:
Actually, now that I look at it, there hasn't been a release in a *long* time (July 2015). I think it's time for a 1.4.0 release. I'll do that in the next day or two.

cweagans:
@pivulic FYI, there is now a 1.4.0 release. Thanks again!

pivulic:
GR8!

Issue #18, opened 2015-09-28
Add a switch to completely disable patching


cweagans:
This is a dependency of https://github.com/cweagans/composer-patches/issues/16 if we move forward there, but it may be useful on its own.

ericduran:
Yea, :+1: this is a good idea.

generalconsensus:
Bump, my composer update is failing because of this

cweagans:
You shouldn't run 'composer update' as part of a ci process, FWIW. I haven't gotten to this yet, but I would merge a PR if you made one.

generalconsensus:
@cweagans What is best practice to be exact?

cweagans:
@generalconsensus 

1. Commit `composer.lock` to the repo
2. When a developer wants to update a composer package, run `composer update your/package` (which will respect the version constraints that are in `composer.json`. Note that you may need to include the `--with-dependencies` flag to allow composer to update a package's dependencies if necessary)
3. In your CI process, when you're producing your build artifacts, you only ever want to run `composer install`, which installs the *exact* versions that are specified in `composer.lock`, rather than any package that happens to match the constraints in `composer.json`. This way, you have a reproducible build.
4. Your `composer.lock` is always considered the canonical place that lists your packages. If it's not in `composer.lock`, it doesn't exist. Note that Composer will complain loudly when your `composer.json` and your `composer.lock` don't match up properly.

That aside, if you install the patches plugin before any other package (i.e. list it first in your `composer.json` `require{}` section), the plugin will detect that you're updating a package that has a patch applied, and will actually *remove* the package from your vendor directory first, and then re-install the package.

phenaproxima:
This functionality is needed to solve the problem in #36, which is a potential security hazard if there is no user interaction. How can we go about implementing this? Can plugins define switches? Should we add it to the extra section in composer.json?

cweagans:
I think something in the extra section in `composer.json` would be okay. I'm thinking something like:

```
"extra": {
    "patches-functionality": true|false
  }
```

or something like that. Feel free to bikeshed the name of the key, but the idea is that there should just be a simple true/false switch to turn on/off the patches functionality.

phenaproxima:
I'd go with ```enable-patching``` or ```enable-patches```. Unless I'm missing something, this seems like a reasonably easy thing to implement...

cweagans:
It should be pretty easy, yeah. Either of those keys sounds good to me.

phenaproxima:
[...]

PR #36, opened 2016-03-25
Dependencies' patches are ignored if the root package does not define any

phenaproxima:
My situation is simple: I have a package that has no patches. It does, however, pull in dependencies that have patches. When I install or update, the dependencies' patches are not applied.

I traced this to gatherPatches() returning early if grabPatches() returns anything falsy. Not sure if this is by design or a bug, but it certainly stops it from working as advertised! :) This here PR should fix the problem.

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -137,7 +137,6 @@ class Patches implements PluginInterface, EventSubscriberInterface {
     $this->patches = $this->grabPatches();
     if ($this->patches == FALSE) {
       $this->io->write('<info>No patches supplied.</info>');
-      return;
     }
 
     // Now add all the patches from dependencies that will be installed.
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -81,6 +81,10 @@ class Patches implements PluginInterface, EventSubscriberInterface {
    * @param Event $event
    */
   public function checkPatches(Event $event) {
+    if (!$this->isPatchingEnabled()) {
+      return;
+    }
+
     try {
       $repositoryManager = $this->composer->getRepositoryManager();
       $localRepository = $repositoryManager->getLocalRepository();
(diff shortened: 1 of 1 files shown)


cweagans:
This PR definitely resolves the problem of not being able to apply dependency patches, however, this is somewhat a security issue. Dependencies can specify patches for any other package, and if that package is being installed, the patch is applied. This can pose a problem if a package author specifies a patch to another package that adds malicious code.

The way I'd prefer to handle this is by fixing #18, but make it so that the switch can explicitly opt-in or opt-out of patching. I think I'd prefer to avoid situations where dependencies can apply patches without the user asking for it (i.e. without explicitly opting in by specifying a patch or by turning on patching via #18), though if you have a strong argument, I could reconsider.

phenaproxima:
I agree with your assessment of the security implications, and I think #18 is a good compromise as long as the user, by opting in, will cause the dependencies' patches to be applied.

phenaproxima:
I have updated the PR to add an ```enable-patching``` switch (enabled by default) so users can explicitly opt out of all patches. Will this do the trick?

phenaproxima:
This PR is a critical release blocker for Lightning, so I'll do whatever I can to get it merged. :)

d70rr3s:
Hi, if i haven't lost myself in translation you say if a define in the root-level an empty "patches" extra key and "enable-patching" key any patches defined by the required packages should be applied, am I right? Well if that is the case isn't working for me. One of my team's member has developed a patch for a contrib module, while its been reviewed/approved we need to apply it to a project in which this contrib module is defined as a dependency for a custom distro we developed. This distro haves its own composer.json file which defines the dependency and the patch to apply. We have added the distro as a dependency for a new project we are working on but the patch is not applied unless we explicitly added to the project composer.json file in the extra/patches section even if the "enable-patching" option is set to true. What am I doing wrong?

phenaproxima:
Hmmm...best I can tell you, you don't need to define an empty patches key in your composer.json. It should work if your root package's composer.json has the following:

```
"require": {
  "cweagans/composer-patches": "^1.5.0"
},
"extra": {
  "enable-patching": true
}
```

d70rr3s:
@phenaproxima thanks I'll test it later and send some feedback

acbramley:
@d70rr3s did you ever get it working? I'm having issues with mine working at all. Adding ```enable-patching: true``` did nothing, when I added an empty "patches" key, I now get an extra line in my composer update output:

"Gathering patches for root package."

But it still then shows

"No patches supplied"

PR #41, opened 2016-04-14
Patch root packages last

deviantintegral:
I'm not sure this is a good idea, but I wanted to put it out for comment.

With enable-patches set to true, I'd expect patches to be applied recursively starting from the furthest dependency, with the root package being last. This PR simply applies root patches last, so if you need to iterate on a dependent patch you can.

Thoughts?

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -142,10 +142,7 @@ class Patches implements PluginInterface, EventSubscriberInterface {
       return;
     }
 
-    $this->patches = $this->grabPatches();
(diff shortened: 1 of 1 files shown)


mwgamble:
Right now if module A has a dependency on module B, and module A has a patch for module B, and they're both being installed at the same time, the patching process will fail because this package attempts to run the patching for module B before module A is installed.

cweagans:
Dependency patching isn't a thing in 2.x. #447 is where most of that is right now.

PR #45, opened 2016-06-08
Document 'enable-patching' attribute.

greg-1-anderson:
Patches stopped working in one of my projects a while back. It appears that now, if you want to use an external patch file, you must enable patching via the "enable-patching" attribute.

This could also be fixed by [adjusting the test for root-level patches](https://github.com/cweagans/composer-patches/blob/master/src/Patches.php#L375), if perhaps that is what was intended.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ Instead of a patches key in your root composer.json, use a patches-file key.
     "preferred-install": "source"
   },
   "extra": {
+    "enable-patching": true,
     "patches-file": "local/path/to/your/composer.patches.json"
   }
 }


cweagans:
Ah, crap, sorry. I was really hoping that change wouldn't break things, but I was in too much of a hurry when it was merged. If there's a patches file in the root, patching should work.

Longer term, this is a good example of where a good suite of tests would really help this plugin. I intend to do this in 2.x unless someone beats me to it.

cweagans:
This field no longer exists in `main`.

Issue #40, opened 2016-04-13
Fail when patches cannot be applied


bartfeenstra:
When a patch cannot be applied, the build simply continues with a warning being output. On automated systems, this is hard to catch. This is a critical issue, as these patches can be used to fix security problems, for instance. Instead, in case of failures, the process should exit with a non-zero error code.

cweagans:
I'd be okay with a flag that controls this. This shouldn't be the default though. Not till 2.0.0 anyway.

bartfeenstra:
I suppose you want the flag for BC reasons, but that does mean this plugin will continue to do the opposite of what it advertises whenever patches fail to apply (which is to apply patches). I'm afraid this will lead to a lot of time being spent by developers trying to find the causes of bugs in their sites or security issues, caused by patches not being applied.

cweagans:
It hasn't been that much of a problem, as the plugin *does* already add some output saying that a patch can't be applied. The main thing is, I don't want to just change a behavior that has been in place since 1.0.0 and hope for the best. A flag is better here because users can opt-in to functionality that will be in 2.0.0 (and in 2.0.0, it'll be an opt-out). I would also not be opposed to some additional messaging at the end of `composer install` or `composer update` if there were any patches that couldn't be applied.

I would also say that before blindly dropping patches into your `composer.json` (*especially* if those are security patches), you should be sure that all the patches will apply in the order that you specify. I'd also say that it's probably much safer to just upgrade the release in the event of a security update, rather than just applying the security patch.

cweagans:
Actually, looks like we already added the flag as an environment variable: https://github.com/cweagans/composer-patches/pull/31

2.0.0 should also clean up the config flags such that they can be specified in composer.json or as an env var, rather than having a weird mix of stuff.

bartfeenstra:
I did not say security releases; I gave examples of problems the current behavior can cause, including those that compromise security.

Adding additional messages will still not solve this problem for automated builds. If a CLI tool does not exit with a non-zero code on failure, it basically fails to tell its executing shell about the failure, which means any non-human observer (build tools) will not be able to take the necessary precautions to prevent worse failures, such as deploying a broken system to production.

cweagans:
That's fair, and I mostly agree with you. However, this still can't be the default behavior until 2.0.0 because semver. It's a BC break. The best we can do in the 1.x releases is more messaging and a flag to let you opt-in to a non-zero exit code.

bartfeenstra:
[...]

PR #46, opened 2016-06-11
Introduce "composer-exit-on-patch-failure" config option.

Leksat:
#40

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -270,7 +270,8 @@ class Patches implements PluginInterface, EventSubscriberInterface {
       }
       catch (\Exception $e) {
         $this->io->write('   <error>Could not apply patch! Skipping.</error>');
-        if (getenv('COMPOSER_EXIT_ON_PATCH_FAILURE') || $this->composer->getConfig()->get('composer-exit-on-patch-failure')) {
+        $extra = $this->composer->getPackage()->getExtra();
+        if (getenv('COMPOSER_EXIT_ON_PATCH_FAILURE') || !empty($extra['composer-exit-on-patch-failure'])) {
           throw new \Exception("Cannot apply patch $description ($url)!");
         }
       }
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -269,11 +269,9 @@ class Patches implements PluginInterface, EventSubscriberInterface {
         $extra['patches_applied'][$description] = $url;
       }
       catch (\Exception $e) {
-        if ($this->composer->getConfig()->get('composer-exit-on-patch-failure')) {
+        $this->io->write('   <error>Could not apply patch! Skipping.</error>');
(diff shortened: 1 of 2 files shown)


cweagans:
This will cause `composer validate` to fail for anyone using this option. I'd prefer not to break this behavior, as some people run `composer validate` as part of their test suite. Having tests suddenly start failing would not be a great thing.

Leksat:
@cweagans, I can't get what you mean... For me `composer validate` works just fine with `"composer-exit-on-patch-failure": true`.

src/Patches.php
```diff
@@ -269,10 +269,12 @@ public function postInstall(PackageEvent $event) {
         $extra['patches_applied'][$description] = $url;
       }
       catch (\Exception $e) {
-        $this->io->write('   <error>Could not apply patch! Skipping.</error>');
-        if (getenv('COMPOSER_EXIT_ON_PATCH_FAILURE')) {
```
bartfeenstra: Removing support for this constant breaks backwards compatibility.

cweagans:
According to the [schema](https://github.com/composer/composer/blob/master/res/composer-schema.json), that should fail. I was going to open a PR against Composer to relax the schema for `config` so that plugins can use it for flags like this, but I can't get `composer validate` to fail now that you mention it. Maybe a bug in the schema validator that Composer uses? I'll investigate more later.

src/Patches.php
cweagans: 👍  ^^ what @bartfeenstra said. Thanks for catching that.

Leksat:
I also restored the `Could not apply patch! Skipping.` message for the case if someone parses the composer output.

src/Patches.php
Leksat: Fixed :wink:

danepowell:
Would the `extra` key be a better alternative to `config`? From the schema docs:

> extra: Arbitrary extra data that can be used by plugins, for example, package of type composer-plugin may have a 'class' key defining an installer class name.

Leksat:
@cweagans, thanks for the explanation. I did not know that `config` options are limited. I moved the option to the `extra` section.

erikhansen:
@Leksat I would love to be able to use this feature. Would you be able to resolve the merge conflict with this PR so that it can be merged in?

dan2k3k4:
Hey @erikhansen, I updated this PR to resolve the conflict. Please test and provide feedback. Thanks in advance!

erikhansen:
@cweagans - Based on my testing, I'd recommend merging this PR.

@dan2k3k4 - I finally made time to test your PR and it's working great on my end. Screenshots demonstrating things working:

*composer.json file*

09-47-08 composer json sites git develop -8iox7

*Exception being thrown and non-zero exit code*
09-45-29 healthyback dev -bash 176x49-loc0b

erikhansen:
@cweagans Any update on this getting included in the next release? Can we expect it in your 2.0 update?

cweagans:
Sorry for the delay, folks.

PR #49, opened 2016-06-15
isPatchingEnabled() checks 'patches-file' too

calin-marian:
isPatchingEnabled() will also take into consideration 'patches-file' when determining if root package uses patches.

I use the patches-file setting in the composer.json file, and after upgrading to 1.5.0 my patches were not applying anymore. I followed the code to the isPatchingEnabled(), where if enable-patching is not set, it will default to true if the root package uses patches, and to false otherwise.

However, detecting if the root project uses patches was done only based on the 'patches' key.

This PR checks also if 'patches-file' exists to determine if the root project uses patches.

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -372,7 +372,7 @@ class Patches implements PluginInterface, EventSubscriberInterface {
   protected function isPatchingEnabled() {
     $extra = $this->composer->getPackage()->getExtra();
 
-    if (empty($extra['patches'])) {
+    if (empty($extra['patches']) && empty($extra['patches-file'])) {
       // The root package has no patches of its own, so only allow patching if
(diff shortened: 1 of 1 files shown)


calin-marian:
#45 is related.

adragus-inviqa:
Wasted a couple of hours because of this. Please merge already.

@calin-marian - care to fix the conflicts?

calin-marian:
I updated the code, no more conflicts.

cweagans:
Sorry for the super long delay on this. I preferred an isset() here, so I fixed this in https://github.com/cweagans/composer-patches/commit/a7e3ebe4d84891b9865d3d7a7120391d079af5eb. This will be included in the next release, which should be out in the next few minutes.

PR #50, opened 2016-07-04
Convert the path to a realpath

h3rj4n:
I was not able to apply local patches so I created this fix, that transforms the relative path to a real path, to enable applying local patches.


cweagans:
Can you provide some information about your environment? OS, Composer version, etc? I want to make sure this doesn't break again in the future. Specifically, 2.x will have a lot of test cases for edge case things like this.

PR #57, opened 2016-07-06
Make local patches work with Git

Leksat:
The Git patch command currently looks like this:

    cd path/to/package && GIT_DIR=. git apply --check -pX path/to/patch/file.patch

As we first `cd` into the package directory, Git is unable to find the patch file at `path/to/package/path/to/patch/file.patch`


cweagans:
Duplicate of #50.

PR #59, opened 2016-07-22
Re-install a package that got removed due to patch changes

### LINKED ISSUE #26: Package deleted but not reinstalled and patched

jurgenhaas:
This should fix #26

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -116,6 +116,9 @@ class Patches implements PluginInterface, EventSubscriberInterface {
             $uninstallOperation = new UninstallOperation($package, 'Removing package so it can be re-installed and re-patched.');
             $this->io->write('<info>Removing package ' . $package_name . ' so that it can be re-installed and re-patched.</info>');
             $installationManager->uninstall($localRepository, $uninstallOperation);
+            $installOperation = new InstallOperation($package, 'Re-install package.');
+            $this->io->write('<info>Re-installing package ' . $package_name . '.</info>');
+            $installationManager->install($localRepository, $installOperation);
           }
         }
       }


mglaman:
I'm noticing this often, which is leaving me with impartial builds; causing us to run `composer update` or `composer install` twice when moving around patches.

grasmash:
This appears to re-install but not re-patch.

grasmash:
#26 is better resolved by #30.

waleedq:
I know that you've dropped support for 1.x, but this issue is still not solved by #30, #30 would solve the re-patch issue for the removed packages, but sometimes the composer is not re-installing the deleted packages for some reason! so i guess this is needed to make sure the package is installed after getting removed.

PR #63, opened 2016-08-03
Issue #26: Package deleted but not reinstalled and patched.

grasmash:


grasmash:
Closing. https://github.com/cweagans/composer-patches/pull/30 resolves the same issue in a better way.

PR #64, opened 2016-08-11
Fix README grammar nit

neclimdul:
I guess "more simple" is not wrong but I feel my high school english teacher giving me a ugly look when I say it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Then your composer.patches.json should look like this:
 
 ## Difference between this and netresearch/composer-patches-plugin
 
-* This plugin is much more simple to use and maintain
+* This plugin is much simpler to use and maintain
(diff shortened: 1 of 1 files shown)


cweagans:
http://english.stackexchange.com/a/2171 says "more simple" is more correct here. I prefer that style because "more $comparative" generally works in most contexts, while "${comparative}er" only works sometimes.

neclimdul:
I actually googled to confirm my memory prior to posting and it proved otherwise but English has terrible grammar inconsistencies so my choice of google phrasing probably lead me to different results.

I also actually considered that "is more simple" flows pretty well, however "is much simpler" seems easier to read than "is much more simple". But as I said both are technically correct so feel free to merge or close at your discretion.

Issue #16, opened 2015-09-02
Patches in packages not installed unless root-level composer.json defines a patches key in extra.


mathieuhelie:
In Patches.php:104 there is a return statement when no patches or patches-file key is found in the root-level extra key of composer. This means that patches that are defined in required packages are not read or applied (this is done at line 107). In order to get patches in required packages to apply, we must add "patches": {} in the extra section of the root-level composer.json file.

If this is desired behavior, an exception should be thrown explaining what is expected.

timmillwood:
I have been having the same issue, it's not ideal. (Will work on a PR).

cweagans:
Honestly, I think I'm okay with this. It avoids the situation where a dependency package requires cweagans/composer-patches and then starts patching any number of other arbitrary projects. Having to add the line in the root composer.json is kind of an "opt-in" for being able to patch things.

Can you describe what circumstances you'd actually want this to happen?

timmillwood:
ok, my use case is:

A Drupal module (currently custom, might be contrib soon) depends on a php library, but there is an open PR for the php library which is needed for this. Therefore looking to add that patch to the PHP library when the module is added (via composer). Therefore the patch plugin won't be defined in the root composer.json, it'd be defined in the module's composer.json.

cweagans:
Right, I understand that part. Let's say some other user wants to use that module and they currently don't have any patches on their project. Is it okay to start mucking around with dependencies on the rest of their site without them knowing it happened?

As it stands right now, they have to opt-in to the patches behavior. If we go down the route of making it on-by-default, we need to have an opt-out switch too. The question is, though, is on-by-default a good choice for something that could potentially introduce security holes into somebody's application? It only takes one malicious library doing that to cause a very large problem for anyone using the lib, so I want to make sure we're covering our bases here.

cweagans:
@mathieuhelie If you have time, I'd like your input on ^ too.

timmillwood:
Ah, that makes sense. Maybe my use case is pretty unique.

cweagans:
I'm not saying no. Let's wait till we have more feedback on it, though.

mathieuhelie:
[...]

PR #70, opened 2016-09-16
Allow dependency patches to be gathered before checking if we should return

acbramley:
This attempts to fix an issue I've encountered which relates to #16

On initially adding a dependency with patches to my site I could get the patches to install, but subsequent patch additions to the dependency wasn't applying them to the root package.

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -93,10 +93,6 @@ class Patches implements PluginInterface, EventSubscriberInterface {
       $packages = $localRepository->getPackages();
 
       $tmp_patches = $this->grabPatches();
-      if ($tmp_patches == FALSE) {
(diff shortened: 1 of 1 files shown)


acbramley:
Hmm, this seems to gather the patches fine, but not actually apply them properly.

acbramley:
Ok, to actually apply the updated patches I did not need to change the Patches.php file. However, the process is a bit of a mess. To apply a new patch from a dependency and make sure the lock file reflected this I had to:

* Remove both the package I'm trying to patch, and the dependency that contains the patch in its composer file
* Run composer install, this means that Patches::gatherPatches will gather patches from the dependency, and be installing the library that is being patched at the same time so Patches::postInstall has access to the patches from the dependency.
* Run composer update --lock

PR #72, opened 2016-09-18
Don't return early on install unless dependencies don't supply patches either

### LINKED ISSUE #71: Patches added to dependency after initial install not applied to root package on update
Steps to reproduce:
- Add a dependency to your root package which contains patches
- Add a new patch to that dependency
- Run composer update my/dependency in your root package

Expected:
New patch is applied

Actual:
New patch is not applied

acbramley:
Fixes #71 

However this may be inefficient, keen to get feedback on this.

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -93,10 +93,6 @@ class Patches implements PluginInterface, EventSubscriberInterface {
       $packages = $localRepository->getPackages();
 
       $tmp_patches = $this->grabPatches();
-      if ($tmp_patches == FALSE) {
-        $this->io->write('<info>No patches supplied.</info>');
-        return;
-      }
 
       foreach ($packages as $package) {
         $extra = $package->getExtra();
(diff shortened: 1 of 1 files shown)


Berdir:
Just noticed this too, the documentation currently claims this is possible but it is actually not. 

Works. I don't see a performance problem, if you get to that point, you already explicitly opted in to patching some way or another.

codebymikey:
This change has been applied on the https://github.com/cweagans/composer-patches/commit/d9d31d812d483cefbae89238e52ba2a25d8a1cbb commit and the PR can be safely closed.

acbramley:
@codebymikey curious why this PR wasn't just merged?

codebymikey:
No idea, I'm not a maintainer of the project.

I just found this PR thinking it would address an issue I was having with patches not applying on initial install.
Then upon attempting to apply the patch, noticed it was already applied a few years back in a different commit, so thought I'd document it here and clean up the list of open PRs.

acbramley:
Haha cheers, didn't realise you weren't a maintainer :)

PR #92, opened 2016-12-20
Fixes #38: Multiple packages using the same patch throws an array to string exception.

### LINKED ISSUE #38: Multiple packages use the same patch throw an array to string exception
If I set up package A with a patch, and package B with the same patch, and set package A to require B, the URLs are merged into an array. This causes composer install to bail with an array to string conversion.

www 2016-04-12 15-21-14

grasmash:


grasmash:
@cweagans Any chance of merging this? It fixes a bug that has been around a while :)

PR #94, opened 2017-01-04
Fix short array syntax in Patches.php:157

echernyavskiy:


cweagans:
Fair warning: version 2 of this plugin will not support PHP 5.3.

Issue #89, opened 2016-12-08
Applying patch on windows failed - tmp directory


lcube45:
I'm having an issue with current configuration :
- windows 10
- composer 1.2.4
- composer-patches 1.6.0

Added patches to my composer.json and when updating or installing, i'm getting the following error :

Patch could not be applied : <patchfile> could not be open, no such file or directory...

After some debugging, problem is in Patches.php line 346 while generating unique id filename...

Adding a real existing path solved the problem for me.

Hope this helps !

ps : i know windows sucks !

lcube45:
Why not using composer cache dir or other global install directory ?

PR #96, opened 2017-01-12
Update Patches.php

b-connect:
Now using system tmp. 

See #89


Leeroy204:
Just tried this out on Windows 7 and it worked!

PR #107, opened 2017-02-26
Re-insert GIT_DIR on non windows platforms

### LINKED ISSUE #106: Removing GIT_DIR=. from patch apply broke functionality on unix when installing dist packages
Before I could install packages from dist and the patches would apply, since PR #104 was merged, this doesn't work. This is clearly described in the documentation:
```
GIT_DIR is the location of the .git folder. If this isn’t specified, Git walks up the directory tree until it gets to ~ or /, looking for a .git directory at every step.
```

daften:
Fixes #106


daften:
@david-garcia-garcia and @celston could you test this on windows please? :)

cweagans:
Closing in favor if #108.

PR #108, opened 2017-02-26
Use --git-dir instead of GIT_DIR env variable

### LINKED ISSUE #106: Removing GIT_DIR=. from patch apply broke functionality on unix when installing dist packages

daften:
Fixes #106 with another method than #107 
This works on unix, but would need to be tested on windows, I'm hoping @david-garcia-garcia and @celston can help with that? :)

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -358,10 +358,10 @@ class Patches implements PluginInterface, EventSubscriberInterface {
     // it might be useful.
     $patch_levels = array('-p1', '-p0', '-p2');
     foreach ($patch_levels as $patch_level) {
-      $checked = $this->executeCommand('cd %s && git apply --check %s %s', $install_path, $patch_level, $filename);
+      $checked = $this->executeCommand('cd %s && git --git-dir=. apply --check %s %s', $install_path, $patch_level, $filename);
       if ($checked) {
         // Apply the first successful style.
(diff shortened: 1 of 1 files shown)


david-garcia-garcia:
I tried the command syntax directly on the console, and it - at least- does not break on windows. Still not sure why we need to use --git-dir at all. We are cd'ing in order to place ourselves in the directory where the git root is supposed to be.

cweagans:
Sometimes there is no git root (i.e. if installing from dist). I prefer this pr over #107, fwiw.

daften:
Yes, if there's no git root inside the package to be patched, this makes sure that's possible.
This is my preferred option too, since it should work cross platform, but it only came to me after I made the other PR :)

dmsmidt:
This should get top priority, without it patching on Linux is broken. 
I tested manually on Linux and it works.

REBELinBLUE:
Any update on this? Patching is completely broken when trying to patch dist packages

REBELinBLUE:
:D Thanks @cweagans

PR #109, opened 2017-02-27
Do not hide command fails.

david-garcia-garcia:
When doing git-apply it will silently fail with stuff such as "skips" because current implementation of executeCommand() does not check error output.

Proposed fix that properly propagates the fact the the command has failed when there is something in the error output.


cweagans:
This is handled differently in #447.

PR #113, opened 2017-03-07
Ignore dependency patches

### LINKED ISSUE #112: Ignoring patches from dependencies doesn't work

```
              }
            }
          }
        }
```

```
```

dmsmidt:
this commit fixes #112

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -180,8 +180,20 @@ class Patches implements PluginInterface, EventSubscriberInterface {
       }
     }
 
-    // Merge installed patches from dependencies that did not receive an update.
+    // Merge installed patches from dependencies that did not receive an update and are not ignored.
     foreach ($this->installedPatches as $patches) {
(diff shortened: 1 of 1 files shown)


dmsmidt:
Note that this only works without the intermediate level (below: drupal/lightning).
I think that abstraction is not really needed and keeping the structure similar to "patches" makes it easier to use.

Before:
```
"patches-ignore": {
      "drupal/lightning": {
        "drupal/panelizer": {
        }
      }
    }
  }
```
After:
```
"patches-ignore": {
  "drupal/panelizer": {
    }
  }
}
```

Note, no cleanup of the code mentioned in #112 is done.

dmsmidt:
Actually this change only sees to work for tag 1.6.0, master doesn't apply patches anyhow.

dmsmidt:
That patches didn't apply is fixed being here in #108. Closing this since we want a PR for master.

PR #118, opened 2017-03-14
Fix ignore dependency patches

### LINKED ISSUE #112: Ignoring patches from dependencies doesn't work

dmsmidt:
This commit fixes #112.

It's a pretty big rewrite.
And I found multiple extra problems that should be solved later (already present before rewrite):
- Patches defined in dependencies are not picked up on a first `composer install`
- Updates of dependencies that include changes to the defined patches are not picked up.

**Note** that this only works without the intermediate level in 'patches-ignore' in composer.json (below: drupal/lightning).
I think that abstraction is not really needed and keeping the structure similar to "patches" makes it easier to use.

Before:
```
"patches-ignore": {
      "drupal/lightning": {
        "drupal/panelizer": {
        }
      }
    }
  }
```
After:
```
"patches-ignore": {
  "drupal/panelizer": {
    }
  }
}
```

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -133,13 +133,15 @@ class Patches implements PluginInterface, EventSubscriberInterface {
           }
         }
 
-        // If we're in verbose mode, list the projects we're going to patch
-        // and the possible patches that are ignored.
-        if ($this->io->isVerbose()) {
-          if (!empty($ignored_patches)) {
-            $this->io->write('<info>Ignore ' . $package_name . ' patches: ' . implode(', ', $ignored_patches) . '</info>');
-          }
+        // If we're in verbose mode, list the patches that are ignored.
+        if ($this->io->isVerbose() && !empty($ignored_patches)) {
+          $this->io->write('<info>Ignore ' . $package_name . ' patches: ' . implode(', ', $ignored_patches) . '</info>');
+        }
+      }
 
+      // If we're in verbose mode, list all found patches per package.
+      if ($this->io->isVerbose()) {
+        foreach ($all_patches as $package_name => $patches) {
           $number = count($all_patches[$package_name]);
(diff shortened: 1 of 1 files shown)


dmsmidt:
This still needs work, since patches defined on dependencies are not picked up untill the dependency is installed. We should also gather patches after everything is installed.

joachim-n:
This PR is really out of date unfortunately.

> Note that this only works without the intermediate level in 'patches-ignore' in composer.json (below: drupal/lightning).
> I think that abstraction is not really needed and keeping the structure similar to "patches" makes it easier to use.

I agree that keeping the structure similar to "patches" is helpful, but:

- changing this at this point breaks semver, because existing 'patches-ignore' values in existing composer.json files will cease to work
- this is an edge case, but it could happen that someone installs two packages that try to specify two versions of a patch that both have the same file name, and the project owner wants to use 'patches-ignore' to ignore just one of them.

frederickjh:
I tested this using the instructions in [first issue that reference this in this comment.](https://github.com/acquia/blt/issues/1379#issuecomment-295868485) This was using the forked and patch repo.

This works but does not work when using a separate patches file. The `patches-ignore` section needs to be in the main composer.json in my test.

I also try going back to an unpatched  version using the syntax in the README with the `patches-ignore` section in the main composer.json but this does not work.

frederickjh:
This pull branch has some issues. At times it will not apply patches and shows no warning and says it found 0 patches, which makes me believe that it maybe did not try. I also had a case where overlapping patches that did not have a conflict also cause all patches to not be applied. 

I have switched to using 1.6.6 instead of this branch as it works for me to apply and ignore patches. If ignore dependencies is working in the releases then I think that this pull request can be closed.

cweagans:
This is too far out of date and too big to do much with. Also not applicable to 2.x (see #447).

PR #120, opened 2017-04-16
Improve patched check

verbruggenalex:
To provide compatibility with https://github.com/Letudiant/composer-shared-package-plugin we should add a check to see if the patch was already applied. Currently if there is a shared package it will create the symlink and try to re-apply the patch. If we first check if the patch was already applied then we will not get a false positive on patch failiure.

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -351,17 +351,21 @@ class Patches implements PluginInterface, EventSubscriberInterface {
       $downloader->copy($hostname, $patch_url, $filename, FALSE);
     }
 
-    // The order here is intentional. p1 is most likely to apply with git apply.
-    // p0 is next likely. p2 is extremely unlikely, but for some special cases,
-    // it might be useful.
-    $patch_levels = array('-p1', '-p0', '-p2');
-    foreach ($patch_levels as $patch_level) {
-      $patched = $this->executeCommand('cd %s && git --git-dir=. apply -R --check %s %s', $install_path, $patch_level, $filename);
-      $checked = $this->executeCommand('cd %s && git --git-dir=. apply --check %s %s', $install_path, $patch_level, $filename);
-      if ($checked && !$patched) {
-        // Apply the first successful style.
-        $patched = $this->executeCommand('cd %s && git --git-dir=. apply %s %s', $install_path, $patch_level, $filename);
-        break;
+    // Check if code is patched already.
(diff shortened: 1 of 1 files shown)


src/Patches.php
```diff
@@ -351,18 +351,21 @@ protected function getAndApplyPatch(RemoteFilesystem $downloader, $install_path,
...
-    // it might be useful.
-    $patch_levels = array('-p1', '-p0', '-p2');
-    foreach ($patch_levels as $patch_level) {
-      $checked = $this->executeCommand('cd %s && git --git-dir=. apply --check %s %s', $install_path, $patch_level, $filename);
-      if ($checked) {
-        // Apply the first successful style.
-        $patched = $this->executeCommand('cd %s && git --git-dir=. apply %s %s', $install_path, $patch_level, $filename);
-        break;
+    // Check if code is patched already.
+    $patched = $this->executeCommand('cd %s && git --git-dir=. apply -R --check -p1 %s', $install_path, $filename);
```
andypost: Why patch level hardcoded?

src/Patches.php
```diff
@@ -351,18 +351,21 @@ protected function getAndApplyPatch(RemoteFilesystem $downloader, $install_path,
...
+
+    if (!$patched) {
+      // The order here is intentional. p1 is most likely to apply with git apply.
+      // p0 is next likely. p2 is extremely unlikely, but for some special cases,
+      // it might be useful.
+      $patch_levels = array('-p1', '-p0', '-p2');
```
andypost: At this point p1 already failed check

src/Patches.php
verbruggenalex: Honestly I dont know the true meaning of patch level. But at that time it seemed important to me not to mess up a reverse check at higher level. And thats the only way I got it to work on my setup. In combination with the shared-packages-plugin.

src/Patches.php
verbruggenalex: I think it was my motivation to just avoid the patching message if it was already patched. I'll see if I can find some time to revisit this PR in the next week.

verbruggenalex: Just noticed why I did not try the other levels. I was too lazy to refractor exiting the loop in order to keep an other level from overriding a valid check. If nobody else needs this (its just an output cleaner) I'll come back to refractor later...

PR #122, opened 2017-04-21
Fix detect dependency patch changes

dmsmidt:
Changes in patch definitions of dependency patches are currently only picked up after a second install or update.

This approach building on fixes of #118 tries to remedy this.

The catch is that we only know after installing/updating dependencies if their composer.lock has been changed. So we need some action in post install/updates.


cweagans:
Patches gathered from dependencies aren't a thing in 2.x. See #447.

PR #124, opened 2017-05-09
Hacky work-around for https://github.com/cweagans/composer-patches/is…

rgristroph:
…sues/123 .

I called this a hacky work-around because it seems the real solution should catch those duplicate entries at creatation, not filter them later.  Also I presume that all the entries in the array are the same and use the first without checking, and the one-line if statement with body is not code standard.  I'm putting up this PR for reference, I may be able to improve it with a real fix later this week.


LionsAd:
Better PR here: https://github.com/cweagans/composer-patches/pull/143/files

PR #129, opened 2017-05-25
Update Patches.php

### LINKED ISSUE #128: Patches do not get applied if root project doesn't have patches
Use case:

I am going to be using drupal core but we are building a distribution which should have all the patches we need to work appropriately.

It seems as though if my root project does not have any patches it doesn't even attempt to look over the package dependencies.

sean-e-dietrich:
Fixes issue where dependency packages are ignore if the root package does not have any patches.

Fixes #128


cweagans:
https://github.com/cweagans/composer-patches#allowing-patches-to-be-applied-from-dependencies

PR #132, opened 2017-05-31
Extra check for uninstallation

lammensj:
We stumbled upon an issue whereby running `composer install --no-scripts` before the regular `composer install` no patches were getting applied. This all due to our deploy strategy with a buid- and deploy-server. 
This makes absolute sense because Patches listens to the install event of a package, which does not occur the second time and is prohibited the first time. Therefor I added an extra check during the checkPatches-function which will search for the patch report. If that does not exists but there are patches available and those are applied (according to composer.lock), then re-install the dependency.

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -431,7 +431,7 @@ class Patches implements PluginInterface, EventSubscriberInterface {
    *
    * @param string $directory
    */
-  function checkPatchReport($directory) {
+  protected function checkPatchReport($directory) {
     return file_exists(sprintf('%s/PATCHES.txt', $directory));
   }
 
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -114,8 +114,10 @@ class Patches implements PluginInterface, EventSubscriberInterface {
           $extra = $package->getExtra();
           $has_patches = isset($tmp_patches[$package_name]);
           $has_applied_patches = isset($extra['patches_applied']);
+          $patch_report_exists = $this->checkPatchReport($installationManager->getInstaller($package->getType())->getInstallPath($package));
           if (($has_patches && !$has_applied_patches)
             || (!$has_patches && $has_applied_patches)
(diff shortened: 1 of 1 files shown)


cweagans:
I'm not immediately opposed to this, but why run `composer install` twice? Why not just run it once with no arguments and call it a day?

lammensj:
This is due to our deployment strategy. On a separate build server, we checkout the project and run `composer install --no-scripts`and `gulp build`. As a final step, the script creates a tarball. At this point, we don't have a connection to a database (among other things) and can not allow composer scripts to be executed.
The next procedure is to download the tarball onto the production server. We run `composer install`again, mainly for the pre- and post-install functionality. composer-patches won't detect unpatched dependencies  because of the .lock file and the already downloaded files.

It was our decision to not run `composer install`on the production server. Imagine what would happen if 3 sites are being deployed on the same shared hosting and composer being run three times. It would take the entire server down.

We tested the adjusted version today and it did exactly what it should be doing: it removes the unpatched dependencies on production so that composer can install them again. The rest is known.

lammensj:
I just ran it in production. Works like a charm :)

steel-track:
I'd like to second merging in this pull request. Our use case is more simple. We don't commit our compiled codebase, but we do commit our composer.lock file. What happens is when we run composer install, the patches_applied line in the lock file causes patching to be skipped because it assumes the packages are already patched. This check for the patch report ensures that the packages are indeed patched.

mrpatrick:
Same scenario for us.  We just tested this fix and it worked like a charm.  Any movement on merging this PR?  Would be super helpful.  Thanks!

nick4fake:
Case is pretty simple:
* COPY composer.jsom
* RUN composer install --no-progress --no-autoloader --no-scripts
* COPY *
* RUN composer install

cweagans:
Writing patch reports isn't a thing in 2.x. #447

Issue #60, opened 2016-07-23
Patch to drupal/core applied twice 


claudiu-cristea:
Drupal core is in `web/`. After applying patch https://www.drupal.org/files/issues/2673628-17.patch (from https://www.drupal.org/node/2673628) the patch correctly applies but I'm getting also a duplicate file in `web/core/b/core/modules/system/src/Tests/Update/CustomStorageEntityDefinitionUpdateTest.php`. I suspect "b" comes from an incorrect usage of system patch command (missing probably the -p1?) but the patch is correctly applied, the file is also present in the right location.

Note that In my case the patch is applied, not from d.o, but from local filesystem.

Relevant composer.json lines:

```
  ...
  "require": {
    "composer/installers": "^1.0.20",
    "cweagans/composer-patches": "~1.0",
    ...
    "drupal/core": "8.1.7",
    ...
  }
  ...
  "extra": {
    "installer-paths": {
      "web/core": ["type:drupal-core"],
      ...
    },
    "patches": {
      "drupal/core": {
        ...
        "Ignore custom storage field definition removals/additions @see https://www.drupal.org/node/2673628": "patches/2673628-17.patch"
      },
      ...
    }
    ...
  }
...
```

wellebee:
I am seeing duplication too, but in two places: `'web/core/b/core'` and `'web/core/core'`.

The following patch directive:

<pre>
"patches": {
    "drupal/core": {
        "node/user reference fields": "https://www.drupal.org/files/issues/2447727-184.patch",
        "audio/video file formatters": "https://www.drupal.org/files/issues/file_field_formatters-1174892-63.patch"
    }
}
</pre>

produces files in `'web/core/b/core/lib'`, `'web/core/b/core/modules'`, `'web/core/core/lib'`, `'web/core/core/modules'`.

cweagans:
This stems from how the plugin tries to figure out the right -p value. I'm not really sure how to make this better, to be honest, but I'll explore some solutions for the 2.x branch.

cweagans:
You'll be able to specify patch depth per-patch in 2.0.0. See #93 for details on that release.

john-herreno-ds:
For current 1.x, [patchLevel](https://github.com/cweagans/composer-patches/tree/1.x#allowing-to-force-the-patch-level--px) should help with patches for drupal/core.

Also, see #178 , #203

cweagans:
[...]

PR #133, opened 2017-06-07
Better deal with core patches if core is installed in a sub directory

mkalkbrenner:
should fix https://github.com/cweagans/composer-patches/issues/60

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -358,31 +358,18 @@ class Patches implements PluginInterface, EventSubscriberInterface {
     // it might be useful. But if the drupal core is installed in a subfolder
     // p2 is the most likely.
     $patch_levels = preg_match('@^[^/\.]+/core$@', trim($install_path, "'")) ? array('-p2', '-p1', '-p0') : array('-p1', '-p0', '-p2');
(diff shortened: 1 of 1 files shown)


hostep:
@mkalkbrenner: please be aware that this module isn't only for patching Drupal projects, it is supposed to be a generic composer plugin for patching all kinds of projects. We are using it for Magento2 projects for example.

Please correct me if I'm wrong, but this PR seems to be targeting Drupal projects, so this change feels very wrong to me...

cweagans:
Yep, what @hostep said. I don't want to add things like this for specific projects.

mkalkbrenner:
I understand that you don't want to have drupal specific stuff.
But the hard-coded order of p1, p0, p2 is really bad if you use https://github.com/drupal-composer/drupal-project for your drupal project.
So what should we do? Maintain a fork of composer-patches?
A different approach would be to have a setting, which options should be tried in which order.
That would also solve https://github.com/cweagans/composer-patches/pull/136

cweagans:
2.x will let you set the p value on a per-patch basis.

PR #134, opened 2017-06-07
optionally prefer patch util over git apply

mkalkbrenner:
In our projects in happens very often that git apply doesn't succeed but the patch utility fallback does.
This PR introduces an prefer-patch-util option for the extras section. If set to true, the patch util is tried before git apply.

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -355,9 +355,8 @@ class Patches implements PluginInterface, EventSubscriberInterface {
     $patched = FALSE;
     // The order here is intentional. p1 is most likely to apply with git apply.
     // p0 is next likely. p2 is extremely unlikely, but for some special cases,
-    // it might be useful. But if the drupal core is installed in a subfolder
-    // p2 is the most likely.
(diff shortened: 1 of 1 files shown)


LionsAd:
+1 looks great to me!

cweagans:
I would prefer to not add a configuration option for this. `git apply` fails, but then `patch` succeeds. Why not just let `git apply` fail in this case? The patch gets applied properly eventually.

mkalkbrenner:
> Why not just let git apply fail in this case?

In our case we handle a lot of patches in our projects. In case that there's a problem with one patch that forces you to switch to the verbose mode you can't see anything because the log if full of error messages. You simply can't find the important part rapidly.
If you can switch the strategy to the one that better fits your project using this PR, the log turns "green" completely, even in verbose mode.
Now, if something goes wrong and you turn on verbose logging you get exactly one "red" part that directly points you to the cause.
That's the story why I created this PR.

As a side effect you have a slight performance benefit.

cweagans:
In 2.x, you can disable the Git patcher or you can write an event subscriber to reorder them. #447

Issue #76, opened 2016-10-05
Apply patch on root folder


tomasnagy:
Is there any way to apply a patch the root of the project, not specifically targeting a package.
Currently I couldn't any documentation about this.
Is this supported in any form or way?

cweagans:
No, there's not. Presumably, those files are committed to your repository. Why not just change them directly?

erikhansen:
@cweagans I too would love to be able to apply patches at the root of a project. My use case for this is as follows: I work on a lot of Magento 2 projects. Magento 2 projects use Composer to install its packages. Unlike applications like Drupal (which seems to use a single Composer package for the entire application) Magento 2 has separate composer packages for its dozens of packages.

There are many situations where a patch would need to be applied to multiple Magento 2 packages simultaneously. With the current state of this plugin, the patch would need to be split between multiple `.patch` files, which would be very clunky. It would be great to be able to apply a patch to multiple packages simultaneously, and it seems like the simplest way to enable this would be to allow patches to be applied to the root of the project. Although allowing patches to be applied to the `vendor` directory would also work, although it seems less elegant.

I'm a contributor to the [capistrano-magento2](https://github.com/davidalger/capistrano-magento2) Gem and we're [currently discussing](https://github.com/davidalger/capistrano-magento2/issues/37#issuecomment-259811943) the best way to patch Magento when deploying it to stage/production environments. In my opinion, using this plugin would be better than building a specific patching mechanism into the capistrano-magento2 Gem. We would just need root-patching support in order for the compser-patches plugin to fully satisfy our requirements.

cweagans:
If you want to code something up to show what this would look like, I'd take a look. I can't promise it'll go in as-is, but I'm certainly not opposed to looking at it.

cweagans:
This will be a part of 2.0, FWIW.

erikhansen:
@cweagans Fantastic news! Any idea of a timeline for 2.0?

cweagans:
When it's done :) I have less than 1 page in my notebook for remaining tasks. This week is crazy, but I have Monday off, so I should be able to take a big bite out of it. I may put that list up on GH later today.

Vedrillan:
+1 

How is the 2.x going ? :)

Trying to apply Magento patches but does not work out of the box if the patch applies to several package at the same time.

erikhansen:
[...]

PR #136, opened 2017-06-07
Add switch to $patch_levels for Magento 2 Official Patches

navarr:
Magento is a fairly popular software where composer-patches is especially useful.  Patch files distributed by Magento in some special cases are based in the root directory, containing paths like `a/vendor/magento/module-catalog`.  In cases such as these, if only a single module is referenced in the patch file then a -p4 flag will allow the patch to work as distributed by Magento.


navarr:
Unnecessary if/when proposal in #76 happens

cweagans:
Works for me. Thanks!

Issue #123, opened 2017-05-02
Applying Patches Drupal8: [ErrorException] Array to string conversion


jiisuominen:
I just updated core to 8.3.1 and updated modules and everything and ran composer update multiple times to get all things updated.

Then after few updates, I started to get this error.

```
Exception trace:
 () at /var/www/drupalvm/vendor/cweagans/composer-patches/src/Patches.php:296
 Composer\Util\ErrorHandler::handle() at /var/www/drupalvm/vendor/cweagans/composer-patches/src/Patches.php:296
 cweagans\Composer\Patches->postInstall() at n/a:n/a
 call_user_func() at phar:///usr/bin/composer/src/Composer/EventDispatcher/EventDispatcher.php:171
 Composer\EventDispatcher\EventDispatcher->doDispatch() at phar:///usr/bin/composer/src/Composer/EventDispatcher/EventDispatcher.php:116
 Composer\EventDispatcher\EventDispatcher->dispatchPackageEvent() at phar:///usr/bin/composer/src/Composer/Installer.php:605
 Composer\Installer->doInstall() at phar:///usr/bin/composer/src/Composer/Installer.php:223
 Composer\Installer->run() at phar:///usr/bin/composer/src/Composer/Command/UpdateCommand.php:158
 Composer\Command\UpdateCommand->execute() at phar:///usr/bin/composer/vendor/symfony/console/Command/Command.php:267
 Symfony\Component\Console\Command\Command->run() at phar:///usr/bin/composer/vendor/symfony/console/Application.php:846
 Symfony\Component\Console\Application->doRunCommand() at phar:///usr/bin/composer/vendor/symfony/console/Application.php:191
 Symfony\Component\Console\Application->doRun() at phar:///usr/bin/composer/src/Composer/Console/Application.php:227
 Composer\Console\Application->doRun() at phar:///usr/bin/composer/vendor/symfony/console/Application.php:122
 Symfony\Component\Console\Application->run() at phar:///usr/bin/composer/src/Composer/Console/Application.php:100
 Composer\Console\Application->run() at phar:///usr/bin/composer/bin/composer:54
 require() at /usr/bin/composer:24
```

I then added some debugging print_r's to Patches.php.

I printed content of that $url variable inside the patching loop @ line 288 of Patches.php.

And this is what I got:
```
Array
(
    [0] => https://www.drupal.org/files/issues/core_path-alias_manager-language_fallback_and_test-2802403-45-D8.patch
[...]

PR #143, opened 2017-07-14
Fix duplicate entries after failed patch.

LionsAd:
This supersedes #142, though installed.json needs to be manually cleaned to get back to a good state.

Closes: #123

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -295,7 +295,6 @@ class Patches implements PluginInterface, EventSubscriberInterface {
       }
       catch (\Exception $e) {
         $this->io->write('   <error>Could not apply patch! Skipping. The error was: ' . $e->getMessage() . '</error>');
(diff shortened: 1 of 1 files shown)


LionsAd:
@cweagans ^^ (this patch is really important!)

cweagans:
Thanks for fixing this! Sorry for the delay. Was AFK.

DrColossos:
I still have the same issue with the patch applied here and described in #142. I replaced all array_marge_recursive() calls with $this->arrayMergeRecursiveDistinct() and all patches apply cleanly every time. It seems like having a distinct structure is the better intention of the code anyways, so we should reuse it

ericchenshine:
The PR 149 fixes the issue. Thanks

DrColossos:
I updated the pull request to the latest master

PR #144, opened 2017-07-27
Multiple packages use the same patch throw an array to string exception

lacun:
Try to complete fix from https://github.com/cweagans/composer-patches/issues/38 when patchs are already installed


boris-brtan:
hello what about this fix?

PR #151, opened 2017-08-30
Don't return too early before gathering patches of dependencies

patkar:
Without this patch, the plugin would already return if there are no patches in the root composer.json. Even with `enable-patching` the early return does not allow to gather patches from dependencies.

Since this is a relative small bugfix, would be cool if you find time to merge that as fast as you can :smirk:  As we want to roll out the patches on a larger project base (>20 projects and counting) :heart: :heart:


patkar:
@cweagans thank you for the merge. Please, may I ask you for a bugfix release too? 😄

cweagans:
https://github.com/cweagans/composer-patches/releases/tag/1.6.2

Issue #148, opened 2017-08-08
Breaks with Git 2.14.0


joergLin:
Using Git 2.14.0 breaks appying Patches (Alpine Edge, Composer 1.5). Latest Working Version is 2.13.4.
```txt
Applying patches for drupal/core
    https://www.drupal.org/files/issues/2551373-35.patch (...)
cd 'web/core' && git --git-dir=. apply --check '-p1' '/tmp/5989b7ae422cc.patch'
warning: unable to access 'config': Is a directory
warning: unable to access 'config': Is a directory
fatal: unknown error occurred while reading the configuration files

cd 'web/core' && git --git-dir=. apply --check '-p0' '/tmp/5989b7ae422cc.patch'
warning: unable to access 'config': Is a directory

warning: unable to access 'config': Is a directory
fatal: unknown error occurred while reading the configuration files

cd 'web/core' && git --git-dir=. apply --check '-p2' '/tmp/5989b7ae422cc.patch'
warning: unable to access 'config': Is a directory
warning: unable to access 'config': Is a directory
fatal: unknown error occurred while reading the configuration files

patch '-p1' --no-backup-if-mismatch -d 'web/core' < '/tmp/5989b7ae422cc.patch'
patch: unrecognized option: no-backup-if-mismatch

BusyBox v1.25.1 (2016-10-26 16:15:20 GMT) multi-call binary.

Usage: patch [OPTIONS] [ORIGFILE [PATCHFILE]]

	-p,--strip N		Strip N leading components from file names
	-i,--input DIFF		Read DIFF instead of stdin
	-R,--reverse		Reverse patch
	-N,--forward		Ignore already applied patches
	-E,--remove-empty-files	Remove output files if they become empty

patch '-p0' --no-backup-if-mismatch -d 'web/core' < '/tmp/5989b7ae422cc.patch'
patch: unrecognized option: no-backup-if-mismatch

BusyBox v1.25.1 (2016-10-26 16:15:20 GMT) multi-call binary.

Usage: patch [OPTIONS] [ORIGFILE [PATCHFILE]]

	-p,--strip N		Strip N leading components from file names
	-i,--input DIFF		Read DIFF instead of stdin
	-R,--reverse		Reverse patch
	-N,--forward		Ignore already applied patches
	-E,--remove-empty-files	Remove output files if they become empty

patch '-p2' --no-backup-if-mismatch -d 'web/core' < '/tmp/5989b7ae422cc.patch'
[...]

PR #152, opened 2017-09-06
Issue #148: Fix --git-dir usage

tstoeckler:
See https://github.com/cweagans/composer-patches/issues/148

See https://stackoverflow.com/questions/1386291/git-git-dir-not-working-as-expected for example


cweagans:
It might be worth replacing the --git-dir and --work-tree args with just -C. I think we'd be able to get rid of the cd in that case as well. Thoughts?

danepowell:
@cweagans I opened a version of this using `-C` here: #165 

I suggest we close this and use #165

PR #153, opened 2017-09-11
Increase post install/update handler priority to support other plugins

mmenozzi:
There are many other Composer's plugins that perform install operation on packages after they have been installed. Many times this operation involves coping packages files elsewhere (look at https://github.com/AydinHassan/magento-core-composer-installer for an example). In such situation it's useful to have patches applied BEFORE the install/copy operation. So, this PR increase the post install/update event handler's priority so other plugins (with priority=0) to run after that patches have been applied.

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -68,12 +68,12 @@ class Patches implements PluginInterface, EventSubscriberInterface {
    */
   public static function getSubscribedEvents() {
     return array(
-      ScriptEvents::PRE_INSTALL_CMD => "checkPatches",
-      ScriptEvents::PRE_UPDATE_CMD => "checkPatches",
-      PackageEvents::PRE_PACKAGE_INSTALL => "gatherPatches",
-      PackageEvents::PRE_PACKAGE_UPDATE => "gatherPatches",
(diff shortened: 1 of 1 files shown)


mmenozzi:
Hi @cweagans,
do you plan to merge this?
If something is not clear let me know.
Thanks.

cweagans:
I don't have an immediate objection to this, but at a minimum, this needs a comment about why the weight is increased so that it doesn't get blown away later.

mmenozzi:
HI @cweagans, you mean a comment in the code?

cweagans:
Yes. Just right above the lines where the weight is increased. Something to the effect of "This is a higher weight for compatibility with [plugin]"

mmenozzi:
@cweagans Ok I'll do it but reopen this pull request please because I'll update the `priority-support` branch 😄

cweagans:
Ah, sorry. Was closing some others. Closed this by mistake.

mmenozzi:
@cweagans done!

cweagans:
Thanks!

mmenozzi:
Thank you to you for the merge!
Do you already know when you'll release a new "stable" version with this added feature?

PR #158, opened 2017-09-26
Allow composer.json to specify failed patch behavior.

### LINKED ISSUE #157: Allow failed patch behavior to be specified in composer.json
Currently an environment variable, or a per-project option is allowed to stop composer if applying a patch fails. It would be nice to be able to specify this globally in the `composer.json` file's config section.

jhedstrom:
- Fixes #157

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -295,7 +295,7 @@ class Patches implements PluginInterface, EventSubscriberInterface {
       }
       catch (\Exception $e) {
         $this->io->write('   <error>Could not apply patch! Skipping. The error was: ' . $e->getMessage() . '</error>');
-        if (getenv('COMPOSER_EXIT_ON_PATCH_FAILURE') || !empty($extra['composer-exit-on-patch-failure'])) {
+        if (getenv('COMPOSER_EXIT_ON_PATCH_FAILURE') || !empty($extra['composer-exit-on-patch-failure']) || !empty($this->composer->getConfig()->get('composer-exit-on-patch-failure'))) {
           throw new \Exception("Cannot apply patch $description ($url)!");
         }
       }


src/Patches.php
```diff
@@ -295,7 +295,7 @@ public function postInstall(PackageEvent $event) {
       }
       catch (\Exception $e) {
         $this->io->write('   <error>Could not apply patch! Skipping. The error was: ' . $e->getMessage() . '</error>');
-        if (getenv('COMPOSER_EXIT_ON_PATCH_FAILURE') || !empty($extra['composer-exit-on-patch-failure'])) {
+        if (getenv('COMPOSER_EXIT_ON_PATCH_FAILURE') || !empty($extra['composer-exit-on-patch-failure']) || !empty($this->composer->getConfig()->get('composer-exit-on-patch-failure'))) {
```
grayside: Looking at this, it seems like maybe the 'extra' line indicates you can configure this in composer.json already in the extra section already?

jhedstrom: I think you're right--I'd mis-read this and assumed it had to be specified per-package.

cweagans: Yep, @grayside is right. Closing, but if that option doesn't work for you, comment here about how it could be better and we'll reopen.

grayside:
@cweagans I tested the extra setting and found it did not result in composer failing with a non-zero exit code.

### composer.json fragment

```json
  "extra": {
    "enable-patching": true,
    "composer-exit-on-patch-failure": 1,
    "patches": {
      "drupal/core": {
        "2699157 - Plugin Lazy loading can cause usort warning":
          "warning.patch"
      }
    },
```

### Using the extra config key

using-extra-config

### Using the environment variable

using-env-config

kdebisschop:
I am seeing the same - environment variable works, settings in extra does not.

kdebisschop:
@cweagans - As noted in my previous comment, I am seeing the same behavior as @grayside. I have concluded that the problem is looking for the extra in $localPackage and have created a PR that makes the setting in composer.json work.

alexdesignworks:
I'd like to emphasise the importance of this. The builds are not failing, but should, when packages got updated and patches failed to apply - as a result - a lot of broken sites after updates because CI passes.

Issue #146, opened 2017-08-04
Install patches from local file


joum:
I created a `.patch` file for a Drupal module that I have been using, and I would like to have it versioned with my codebase, but not publicly online. That being said, I would like to have the file stored locally and use a similar syntax to what is already possible, but referring a local file instead of an URL. Is this possible?

cweagans:
Yes, this is possible. Instead of a URL, you can specify a path to your patch file relative to the directory that your composer.json is in. For instance, if the patch file is in the same directory as your composer.json, the URL would be either 'patchname.patch' or './patchname.patch'.

cweagans:
(For reference, the plugin just uses file_get_contents() under the hood, so anything you can pass as the first argument will be valid for this plugin)

joum:
@cweagans Thank you for your reply. Is it possible that some folder permissions interfere with this process? I created a `patches` directory next to my `composer.json` file and then I referenced a `.patch` file like you mentioned above and composer terminated with a file not found exception.

I tried debugging your module for a bit and found out it was breaking in the `getAndApplyPatch` function. The `eventDispatcher`triggered for every patch I had, but then the `file_exists` never got the real filename, and then advanced to the download option, which in turn failed because it isn't a valid URI.

 I eventually found another way to solve my problem, but if it's helpful I could open an issue with more details and the error messages.

cweagans:
If you can post a composer.json that reproduces the problem here, I can take a look.

milkovsky:
Change your patch extension to `txt`. Example:

```
"extra": {
        "patches": {
            "drupal/views_infinite_scroll": {
                "Integrate with Ajax Assets Plus" : "patches/foo-bar.patch.txt"
            }
        }
    }
```

marklabrecque:
Adding .txt is what fixed it for me. Thanks @milkovsky !

uberhacker:
The `.txt` extension trick didn't work for me.  Why would an extension matter?

uberhacker:
@cweagans: Can this issue be reopened?  The current version is not accepting local patches named `.patch` or `.txt`.

marklabrecque:
Still works fine for me. Please ensure your path is setup correctly to reference from the project root for composer to find the file in question

uberhacker:
I don't get a file not found message.  It just doesn't apply.  FWIW, this is a Drupal core patch.  I still haven't received an answer to my question about the `.txt` extension.  That doesn't make any sense.

[...]

PR #161, opened 2017-10-04
Correctly resolve relative local paths pointing to patch files from dependencies

smuuf:
This ***may be quite naive solution*** and may need some more work *(it may  break something else elsewhere)*, but this already solved my problem where - *having this multi-project structure* - ...

```
- MY_PROJECT_1
  - Requires: MY_PROJECT_2
  - Extra:
    {
      "enable-patching": true,
      "patches": {}
    },
- MY_PROJECT_2
  - Requires: 3RD_PARTY_PROJECT
  - Extra:
    {
      "patches": {
        "3RD_PARTY_PROJECT": {
          "Important patch": "./patches/an_important.patch" // Path relative to MY_PROJECT_2's root.
        }
    }
  
```

... the `./patches/an_important.patch` specified relatively to *MY_PROJECT_2*'s path ***wouldn't be properly resolved*** when running `composer install/update` inside the *MY_PROJECT_1*'s root directory.

> The "./patches/an_important.patch" file could not be downloaded: failed to open stream: No such file or directory

It seems to be solved by first trying whether the `dependency's root path` + `relatively specified path to dependency's patch` does - by any chance - actually represent an existing path.

If it does, that new absolute path is used instead. This IMO should handle most of the cases where the path to the patch file is specified some other way - an *URL*, some *absolute path* or *deliberately specified relative path*. In these cases `realpath()` should return false and originally specified path should be used.

*Any thoughts?*

---

Also, may be related to https://github.com/cweagans/composer-patches/issues/146.

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -102,6 +102,16 @@ class Patches implements PluginInterface, EventSubscriberInterface {
       foreach ($packages as $package) {
         $extra = $package->getExtra();
         if (isset($extra['patches'])) {
+          // Go through each of this package's defined patches and make sure paths to patches
+          // defined relative to their home package are resolved properly.
+          $packageDir = $installationManager->getInstallPath($package);
+          foreach ($extra['patches'] as &$singlePackage) {
+            foreach ($singlePackage as &$path) {
+              if (realpath($tmp = "$packageDir/$path")) {
+                $path = $tmp;
+              }
+            }
+          }
           $this->installedPatches[$package->getName()] = $extra['patches'];
         }
         $patches = isset($extra['patches']) ? $extra['patches'] : array();


dsnopek:
This functionality would be awesome!

rlmumford:
I would love this feature! Not sure what the conflicts are though

ShirishRam:
This would be a very good feature to have. Nested dependencies are commonly used by many.

cweagans:
Sorry for the slow response - I'm in the process of moving.

On one hand, I see where this would be useful, but on the other, I could see this getting abused by e.g. Drupal contrib modules that want to patch core. Those core patches should really be in an issue queue somewhere, and by extension should have an actual URL that can be used. I'm not really sure how I feel about this.

rlmumford:
No worries!

I understand you point about potential abuses of this feature but let me lay out my use case. I maintain package that basically lists all the drupal modules that I use on a regular basis and any changes to them that I need to make. Periodically, there is a patch that I work on for a given contrib module, that the module maintainer chooses not to pursue the request or to develop in a different way (that isn't compatible with the problem I was trying to solve) in the mean time, I'm trying to maintain a collection of high traffic websites. Being able to maintain my own set of patches is the way to keep this sane.

The specific problem that lead me to develop this PR was with the "paragraphs" drupal module. Currently the "paragraph" content does not get included REST json output. I submitted a patch to paragraphs module to rectify this. The module maintainer decided to try and fix the problem upstream (entity reference revisions) but this has now dragged on for > 1 year AND the solution being developed is not compatible with my initial design. Swapping to the upstream (still unfinished) implementation would mean re-configuring 7 different applications with the potential of having to do it again if the upstream implementation changes again.

What I would prefer to do in this case, is simply maintain my own patch for the paragraphs module and make sure it applies with each new release of paragraphs.

rlmumford:
(my work is in pr #218

PR #165, opened 2017-11-08
Fixes #148: Incompatible with Git 2.14+.

### LINKED ISSUE #148: Breaks with Git 2.14.0

danepowell:


danepowell:
Ironically, while this does solve the errors thrown by Git 2.14, it actually triggers another bug, and patches don't get applied correctly.

Back when `git apply` was failing, composer-patches would fall back to using the patch command, i.e.:
`patch '-p2' --no-backup-if-mismatch -d 'docroot/core' < '/tmp/5a0340574777b.patch'`

This worked great. But for some reason, now that it's using `git apply`, it fails silently.

danepowell:
Debugging with this command:
`set -x; GIT_TRACE=2 GIT_CURL_VERBOSE=2 GIT_TRACE_PERFORMANCE=2 GIT_TRACE_PACK_ACCESS=2 GIT_TRACE_PACKET=2 GIT_TRACE_PACKFILE=2 GIT_TRACE_SETUP=2 GIT_TRACE_SHALLOW=2  git -C 'docroot/core' apply '-p2' /home/dane/sites/blted89/2883813-27.patch -v -v; set +x`

I see this:
```
+ GIT_TRACE=2
+ GIT_CURL_VERBOSE=2
+ GIT_TRACE_PERFORMANCE=2
+ GIT_TRACE_PACK_ACCESS=2
+ GIT_TRACE_PACKET=2
+ GIT_TRACE_PACKFILE=2
+ GIT_TRACE_SETUP=2
+ GIT_TRACE_SHALLOW=2
+ git -C docroot/core apply -p2 /home/dane/sites/blted89/2883813-27.patch -v -v
09:58:15.310932 trace.c:333             setup: git_dir: .git
09:58:15.310973 trace.c:334             setup: git_common_dir: .git
09:58:15.310979 trace.c:335             setup: worktree: /home/dane/sites/blted89
09:58:15.310984 trace.c:336             setup: cwd: /home/dane/sites/blted89
09:58:15.310988 trace.c:337             setup: prefix: docroot/core/
09:58:15.310993 git.c:344               trace: built-in: git 'apply' '-p2' '/home/dane/sites/blted89/2883813-27.patch' '-v' '-v'
Skipped patch 'modules/hal/tests/src/Functional/EntityResource/Media/MediaHalJsonAnonTest.php'.
Skipped patch 'modules/media/tests/modules/media_test_type/media_test_type.info.yml'.
Skipped patch 'modules/media/tests/src/Functional/MediaFunctionalTestCreateMediaTypeTrait.php'.
Skipped patch 'modules/media/tests/src/Functional/MediaRevisionTest.php'.
Skipped patch 'modules/media/tests/src/Functional/MediaSourceFileTest.php'.
Skipped patch 'modules/media/tests/src/Functional/MediaUiFunctionalTest.php'.
Skipped patch 'modules/media/tests/src/FunctionalJavascript/MediaSourceFileTest.php'.
Skipped patch 'modules/media/tests/src/FunctionalJavascript/MediaSourceImageTest.php'.
Skipped patch 'modules/media/tests/src/FunctionalJavascript/MediaUiJavascriptTest.php'.
Skipped patch 'modules/media/tests/src/FunctionalJavascript/MediaViewsWizardTest.php'.
Skipped patch 'modules/media/tests/src/Kernel/MediaSourceFileTest.php'.
Skipped patch 'modules/rest/tests/src/Functional/EntityResource/Media/MediaResourceTestBase.php'.
Skipped patch 'modules/media/config/install/core.entity_form_display.media.file.default.yml => profiles/standard/config/optional/core.entity_form_display.media.file.default.yml'.
Skipped patch 'modules/media/config/install/core.entity_form_display.media.image.default.yml => profiles/standard/config/optional/core.entity_form_display.media.image.default.yml'.
Skipped patch 'modules/media/config/install/core.entity_view_display.media.file.default.yml => profiles/standard/config/optional/core.entity_view_display.media.file.default.yml'.
Skipped patch 'modules/media/config/install/core.entity_view_display.media.image.default.yml => profiles/standard/config/optional/core.entity_view_display.media.image.default.yml'.
Skipped patch 'modules/media/config/install/field.field.media.file.field_media_file.yml => profiles/standard/config/optional/field.field.media.file.field_media_file.yml'.
Skipped patch 'modules/media/config/install/field.field.media.image.field_media_image.yml => profiles/standard/config/optional/field.field.media.image.field_media_image.yml'.
Skipped patch 'modules/media/config/install/field.storage.media.field_media_file.yml => profiles/standard/config/optional/field.storage.media.field_media_file.yml'.
Skipped patch 'modules/media/config/install/field.storage.media.field_media_image.yml => profiles/standard/config/optional/field.storage.media.field_media_image.yml'.
Skipped patch 'modules/media/config/install/media.type.file.yml => profiles/standard/config/optional/media.type.file.yml'.
Skipped patch 'modules/media/config/install/media.type.image.yml => profiles/standard/config/optional/media.type.image.yml'.
09:58:15.311544 trace.c:435             performance: 0.000773097 s: git command: 'git' '-C' 'docroot/core' 'apply' '-p2' '/home/dane/sites/blted89/2883813-27.patch' '-v' '-v'
+ set +x
```

Git is silently skipping those file renames. No idea why. I've tried that command with different patch levels (p0, p1...) as well, none of them work.

Could it be because `docroot/core` was checked out as a distributed package, rather than from source? So it's not a git repo, and `git apply` just ignores it.

danepowell:
This would explain it: https://stackoverflow.com/a/27283285

cweagans:
> Could it be because docroot/core was checked out as a distributed package, rather than from source? So it's not a git repo, and git apply just ignores it.

I think `git apply` doesn't work on non git repos, which is why we have the `patch` fallback.

I think we should probably solve one thing at a time. If this change resolves things for the new version of Git (ideally without breaking earlier versions), we should move forward here, and then address the other issue (patches sporadically not getting applied) separately.

danepowell:
Okay, I added a conditional that attempts to catch instances where `git apply` silently fails (such as if you are running it against a distribution of a package, but your application as a whole is under version control).

I think this is the best way to preserve the current business logic while fixing the incompatibilities with newer versions of Git.

Long-term, I think we should stop using `git apply` altogether, or apply the patches in a temporary directory outside of version control and then move them into place.

andypost:
> I think git apply doesn't work on non git repos, which is why we have the patch fallback
Btw why `patch` is fallback, it looks it should depend on ”prefer stable" flag or maybe use composer api to detect source of package.

Otoh patch is most universal way so git apply should be fallback

danepowell:
This should be good to merge. I'm not really sure how to increase coverage in this instance to get that test to pass.

hostep:
There is also the `--directory` option of `git apply` which might help here.
(please ignore my comment if irrelevant, I just quickly read through this thread)

cweagans:
> Btw why patch is fallback, it looks it should depend on ”prefer stable" flag or maybe use composer api to detect source of package.

@andypost The plugin should use the vcs-appropriate way first (right now, only git is supported, but there's no reason we can't support svn and similar), then `patch` as the fallback. `git` is the default because pretty much everyone has that installed. Fewer people have `patch` installed.

danepowell:
@cweagans I think this is good to merge, is there anything else I can do to improve it?

cweagans:
This looks good. If you make a PR against 1.x, I'll merge that and tag a 1.x release as well.

RalfEisler:
Patches still fail under the following conditions:

"cweagans/composer-patches": "^1.6.4",
git version 2.15.1
patch 2.0-12u10 FreeBSD

Error messages:
Could not apply patch! Skipping. The error was: Cannot apply patch https://www.drupal.org/files/issues/1356276-408--8.4.3.patch
etc.

danepowell:
@RalfEisler let's not have a discussion in two places, if you can provide a minimum example to reproduce please open a new issue. But as I commented in #148, I suspect there's something with that particular patch, not with Composer.

PR #167, opened 2017-11-11
Fixes #168: skip duplicate patches

### LINKED ISSUE #168: Fails when a package is called in twice with the same patch
Given a structure like this:

```
Top Level
  Pacakge 1
    Subpackage A
    Subpackage B
        Patch
    Subpackage C
  Package 2
     Subpackage B
        Patch
     Subpackage D
```

The patch will get queued twice and on the second application will fail. You can also see this if you apply the same patch URL with a different description in one file:

```
{
  "name": "cweagans/composer-patches-test-project",
  "description": "Project for use in cweagans/composer-patches acceptance tests.",
  "type": "project",
  "license": "BSD-2-Clause",
  "repositories": [
    {
      "type": "path",
      "url": "../composer-patches"
    }
  ],
  "require": {
    "cweagans/composer-patches": "*@dev",
    "drupal/drupal": "~8.2"
  },
  "extra": {
    "patches": {
      "drupal/drupal": {
        "Add a startup config for the PHP web server": "https://www.drupal.org/files/issues/add_a_startup-1543858-58.patch",
        "Add a startup config for the PHP web server again": "https://www.drupal.org/files/issues/add_a_startup-1543858-58.patch"
      }
    }
  }
}
```

I think if the same patch URL is presented twice, it should only be applied the first time.

kdebisschop:
When the same patch is called in twice, it will generate a conflict on the second apply.

> $ composer update
> Gathering patches for root package.
> Removing package drupal/coder so that it can be re-installed and re-patched.
>   - Removing drupal/coder (8.2.12)
> Loading composer repositories with package information
> Updating dependencies (including require-dev)                              
> Package operations: 1 install, 0 updates, 0 removals
> Gathering patches for root package.
> Gathering patches for dependencies. This might take a minute.
>  - Installing drupal/coder (8.2.12): Cloning 984c54a7b1 from cache
>  - Applying patches for drupal/coder
>    https://www.drupal.org/files/issues/2804739-7.patch (Allow underscore properties in Annotation classes)
>    https://www.drupal.org/files/issues/2804739-7.patch (Allow underscore properties in Annotation class)
>   Could not apply patch! Skipping. The error was: Cannot apply patch https://www.drupal.org/files/issues/2804739-7.patch

You can see the effect if you add a patch to a composer.json file twice. Of course, you can say that is just a typo. The problem really is the case where two child packages bring in the same patch -- if the system collects patches from lower-level composer.json files, the particular patches that are applied are much harder to control over time.

This patch simply skips patches with the same URL on the second attempt to apply them. It should be possible to accomplish the same effect when gathering patches, but that only occurred to me later and I was not able to see as simple a fix there. Perhaps arrayMergeRecursiveDistinct() needs to consider value instead of or as well as key -- but I was not entirely sure about that operation so I opted for this strategy.

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -307,7 +307,7 @@ class Patches implements PluginInterface, EventSubscriberInterface
         $extra['patches_applied'] = array();
 
         foreach ($this->patches[$package_name] as $description => $url) {
-            if (in_array($extra['patches_applied'], $url, true)) {
+            if (in_array($url, $extra['patches_applied'], true)) {
                 continue;
             }
             $this->io->write('    <info>' . $url . '</info> (<comment>' . $description . '</comment>)');
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -307,6 +307,9 @@ class Patches implements PluginInterface, EventSubscriberInterface
         $extra['patches_applied'] = array();
 
         foreach ($this->patches[$package_name] as $description => $url) {
+            if (in_array($extra['patches_applied'], $url, true)) {
+                continue;
+            }
             $this->io->write('    <info>' . $url . '</info> (<comment>' . $description . '</comment>)');
             try {
                 $this->eventDispatcher->dispatch(


cweagans:
Gathering patches from dependencies isn't a thing in 2.x, so this will be less of a problem. #447

PR #175, opened 2017-12-01
Fixes #174: Compatibility with older Git versions.

### LINKED ISSUE #174: Patches fail to apply with older versions of Git
With Composer Patches 1.6.3, patches will silently fail to apply on machines with older versions of Git installed. I _think_ it will fail for versions prior to Git 2.9.0, based on this commit: https://github.com/git/git/commit/3f5794493c7a0ba2bba0278b9129d98df343d341

Empirically, I've tested and verified that Git versions 2.11.0 and above work fine, while 2.7.4 and before fail. I haven't had time to build the versions in between from source in order to test.

I haven't yet determined whether this is a regression introduced by https://github.com/cweagans/composer-patches/pull/165

I don't see why that would have caused a regression. But in testing, Composer Patches 1.6.2 works fine while Composer Patches 1.6.3 fails in at least one environment: https://github.com/acquia/lightning/pull/524

I think the problem is that older versions of `git apply` silently fail to apply patches, so there's no way for Composer Patches to verify whether they've applied correctly or not. But again, I'm not sure how to reconcile that with the changes in https://github.com/acquia/lightning/pull/524, which should have only been more aggressive in catching `git apply` failures.

danepowell:
I'm still testing this locally but wanted to get this out so we could start review.

The first commit is the only functional change: adding a conditional to only use `git apply` on git repos.

The second commit is a refactor to encapsulate this functionality and make it a little cleaner to edit in the future.

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -361,37 +361,13 @@ class Patches implements PluginInterface, EventSubscriberInterface {
       $downloader->copy($hostname, $patch_url, $filename, FALSE);
     }
 
-    // Modified from drush6:make.project.inc
-    $patched = FALSE;
+    // The order here is intentional. p1 is most likely to apply with git apply.
+    // p0 is next likely. p2 is extremely unlikely, but for some special cases,
+    // it might be useful. p4 is useful for Magento 2 patches
+    $patch_levels = array('-p1', '-p0', '-p2', '-p4');
 
-    // Do not use git apply unless the install path is itself a git repo
-    // @see https://stackoverflow.com/a/27283285
-    if (is_dir($install_path . '/.git')) {
-      // The order here is intentional. p1 is most likely to apply with git apply.
-      // p0 is next likely. p2 is extremely unlikely, but for some special cases,
-      // it might be useful. p4 is useful for Magento 2 patches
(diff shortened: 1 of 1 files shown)


danepowell:
I've confirmed that this works with both Git 2.1.4 and Git 2.14.1, the oldest and newest releases that you'll typically see in the wild.

@cweagans if this looks good to you and you can merge it, I'll try to test the dev branch in more places, and then maybe we can tag a new release

danepowell:
@cweagans I've tested and verified this fix in all of our environments that were failing with 1.6.3. Would you mind merging and tagging a new release? Judging by the comments in #174, this is biting a whole lot of people.

benjaminrau:
@cweagans @danepowell One should check github issues first, then debug issues oneself ;)
I debugged my scrutinizer build for few hours now... finally i solved it by just requiring this patch.

Read: Tested it right now and it works fine!

pwolanin:
Thanks, this fixed the issue for me with git version 2.7.4.

under the covers it's falling back to the patch command with this change

cweagans:
Sorry for the slowness here folks.

PR #176, opened 2017-12-01
patches-dir feature for generating patch definitions from filesystem structure

eerorika:
If a project has more than a few patches, they should be stored in a directory to keep them maintainable. The filesystem structure can serve as an implicit "patches" array which declares the patches. This allows for simpler configuration.

#### Usage

1. Store patches in a directory structure `$patches_dir/$vendor/$package/$patchfile`
Example:

```
patches
├── magento
│   ├── framework
│   │   ├── descriptive-patch-name.patch
│   │   └── another.patch
│   ├── magento2-base
│   │   ├── more.patch
│   │   └── and-so-on.patch
├── mageplaza
│   └── magento-2-seo-extension
│       └── many-many-patches.patch
```

2. Point to that directory with `patches-dir` setting in `composer.json`

```json
{
    "extra": {
        "patches-dir": "patches",
    }
}
```

---

`patches-dir` can be used in conjunction with `patches` or `patches-file`. Patch declarations generated from `patches-dir` are merged with those from `patches` or `patches-file` because `patches-dir` is only usable with locally stored patches, and a project might need externally downloaded patches as well.

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -225,10 +225,10 @@ class Patches implements PluginInterface, EventSubscriberInterface
     {
         // First, try to get the patches from the root composer.json.
         $extra = $this->composer->getPackage()->getExtra();
+        $patches = array();
         if (isset($extra['patches'])) {
             $this->io->write('<info>Gathering patches for root package.</info>');
             $patches = $extra['patches'];
-            return $patches;
         } // If it's not specified there, look for a patches-file definition.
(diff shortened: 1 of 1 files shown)


cweagans:
I'd rather not do this. #447 leans further into defining patches explicitly (including per-patch depth settings + sha256 hashes for file integrity). It does leave the door open to writing your own patch resolver plugin that would let you do this if you'd like.

Issue #178, opened 2017-12-13
Use dry-run to test patch in order to avoid empty folders being generated


ericchenshine:
Currently, I am trying to upgrade Drupal from 8.3.7 to 8.4.3 including BLT, Lightning, and Drush upgrade. I have found there are some bizarre files have been generated on the air after each build.

As you can see the core dir appears in 3 places as below:

![wrong files generated by patch process](https://user-images.githubusercontent.com/26533900/33918055-8fcab422-e005-11e7-9885-a9a526415e87.png)

- docroot/core
- docroot/core/core
- docroot/core/b/core

This does not smell right and after investigation, I have found that those unwanted files are generated after patches defined in lightning has applied to `drupal/core`. I have checked this issue on our dev, test, and prod environments on Acquia Cloud and have found that this problem also happens on those servers. So this issue happens prior to my upgrade, I just happen to notice that during the upgrade work.

**Here are the details of my local environment:**

- drush: 8.1.15
- patch: 2.5.8
- composer: 1.5.5
- cweagans/composer-patches: 1.6.4

**Here is the log with my debug messages:**

```
Gathering patches for root package.
Warning: The value for extras.installer-paths in composer.json differs from BLT's recommended values.
See https://github.com/acquia/blt/blob/8.x/template/composer.json
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Package operations: 1 install, 0 updates, 0 removals
Gathering patches for root package.
Gathering patches for dependencies. This might take a minute.
  - Installing drupal/core (8.4.3): Loading from cache
  - Applying patches for drupal/core
    https://www.drupal.org/files/issues/2752961-90.patch (Clear Twig caches on deploys)
$install_path: docroot/core
  - Applying patches for https://www.drupal.org/files/issues/2752961-90.patch
$patch_level: -p1
$patch_level: -p0
$patch_level: -p2
    https://www.drupal.org/files/issues/ignore_front_end_vendor-2329453-116.patch (Ignore front end vendor folders to improve directory search performance)
$install_path: docroot/core
  - Applying patches for https://www.drupal.org/files/issues/ignore_front_end_vendor-2329453-116.patch
$patch_level: -p1
$patch_level: -p0
$patch_level: -p2
    https://www.drupal.org/files/issues/1356276-408--8.4.3.patch (1356276 - Allow profiles to provide a base/parent profile and load them in the correct order)
$install_path: docroot/core
[...]

PR #179, opened 2017-12-13
Use 'dry run' to test patch first (against branch 1.x)

ericchenshine:
This is related to the issue #178 .

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -362,8 +362,8 @@ class Patches implements PluginInterface, EventSubscriberInterface {
     }
 
     // The order here is intentional. p1 is most likely to apply with git apply.
-    // p0 is next likely. p2 is extremely unlikely, but for some special cases,
-    // it might be useful. p4 is useful for Magento 2 patches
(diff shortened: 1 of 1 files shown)


ericchenshine:
The patch is originally created when I was using Patch version 2.5.8 on Mac OS. However, after I installed GNU Patch version 2.7.1, this PR does not work anymore. Even using dry-run, empty folders will be still created.

waleedq:
Hello @ericchenshine, thank you for this pull request it helped me to fix an issue with my project as the patches were getting applied partially because the script is not testing the patch before, and that's right for 2.7.1 the --dry-run is not returning any errors after execution.

To fix that you only need to use the --posix flag or --no-backup-if-mismatch along with -f to force the test without asking any questions or assuming reverse patches, you can check the full command here 
https://github.com/waleedq/composer-patches/blob/1.7.1/src/Patches.php#L397

They might have dropped the support for 1.x that's why i didn't create a new pull request, just for anyone who's facing the issue with partial patches please use this pull with -f --posix, or you can use the 1.7.1 tag from my repo.

PR #188, opened 2018-01-16
Support for OSX and Linux testing with Travis

Techwolf12:
Test if travis will build linux and OSX at the same time

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,10 @@ php:
   - '7.1'
   - nightly
 
+os:
+  - linux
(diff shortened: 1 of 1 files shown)


cweagans:
Thanks for this! As I mentioned in the other issue, there's something strange happening with Travis today, so going to hold off on this until Travis is happy.

Worst case, I'll switch everything over to Circle CI, which also supports Mac and Linux.

cweagans:
We don't use travis anymore.

PR #189, opened 2018-01-17
1.x backport: Allows patch-level to be specified per-project.

### LINKED ISSUE #43: Patch  for new file  applied to incorrect location
https://www.drupal.org/files/issues/drupal-migrate_skip_on_value-2711949-2.patch

I'm using drupal-composer/drupal-project and when I try to apply the patch listed above, it creates it as web/core/core/core/modules/migrate/src/Plugin/migrate/process/SkipOnValue.php, etc.

jhedstrom:
- Backport to `1.x`
- Reroll of #101
- Fixes #43

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -300,7 +300,7 @@ class Patches implements PluginInterface, EventSubscriberInterface {
       $this->io->write('    <info>' . $url . '</info> (<comment>' . $description. '</comment>)');
       try {
         $this->eventDispatcher->dispatch(NULL, new PatchEvent(PatchEvents::PRE_PATCH_APPLY, $package, $url, $description));
-        $this->getAndApplyPatch($downloader, $install_path, $url);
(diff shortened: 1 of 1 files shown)


bkosborne:
Would love to see this in! We're also having issues with applying Drupal core patches that introduce new files. They go to the wrong place.

MPParsley:
Indeed, this should go in 1.x

fenstrat:
This works a treat for core patches. Would also love to see this in 1.x.

cweagans:
Despite what I said before, I'm okay with this. It's pretty low risk and solves some immediate problems for people using composer-patches. While I would really like to say that this is a 2.x feature, the reality is that 2.x isn't moving along as quickly as I'd like, and it's not fair to make people wait for something that doesn't have a realistic release date set.

Thanks for bringing this back up, and sorry I didn't give it more consideration before.

PR #184, opened 2017-12-19
Fix a Unix breaking bug caused by a hack

Techwolf12:
~~This first checks if you are running windows before changing the patch command to a flag not supported on FreeBSD.~~
Patch is given the --posix flag now.

Should be a fix for #182 
If someone with access to multiple windows machines can test this a bit more thoroughly, would be appreciated!

PR #190, opened 2018-01-25
1.x Backport: Change patch behaviour to be posix instead of a flag that breaks unix

Techwolf12:
Same one as #184 but for the 1.x branch


joergLin:
Thanks! Had to install "patch" on Alpine Linux, otherwise I get `patch: unrecognized option: posix`. Was working without it before.

cweagans:
#447

PR #195, opened 2018-02-08
Fixed incorrect path issue with relative paths when applied from dependencies

InfopactMLoos:
When applying patches from dependencies the relative paths are used from the root of the project. These  would of course need to be relative to the dependency's path (ie. ./vendor/author/project). To fix this I have added some checks to discover if the path is relative and if so turn the path into a absolute path to prevent any path mistakes regardless of current directory.


cweagans:
Gathering patches from dependencies is not a thing in 2.x, so applying patches stored relative to the root of a dependency isn't either

Issue #182, opened 2017-12-19
Windows fixing "hack" actually breaks FreeBSD (and Unix)


Techwolf12:
Hi!

The hack in this file: https://github.com/cweagans/composer-patches/blob/3466175168856f843ca3c0c05b3563633d226526/src/Patches.php#L426

Breaks support for BSD patch (commonly found under FreeBSD servers or some Mac devices, in my case version: patch 2.0-12u11 FreeBSD)
`--no-backup-if-mismatch` is the flag that breaks this. 

My suggestion is a check for runtime (if running on windows or using gnupatch, add the flag, else leave it out).
I will check if I can create a pull request within a month. Feel free to do this earlier.

drupol:
Using FreeBSD and having the same issue.

Removing the faulty flag fixed the issue.

criley:
Any word on this getting fixed? I have a major project that I was going to convert to Drupal 8 however this really puts a damper on it.

kimpepper:
I came across this issue using Alpine Linux. The fix was to explicitly install the patch package which bumps up the version from 1.x to 2.x.

drupol:
@kimpepper could you describe how you did to fix this ? I'm unable to get it working here on FreeBSD.

Thanks!

kimpepper:
Just `apk add -u patch` on alpine. This bumped the default version of the `patch` package.

drupol:
@cweagans The issue with FreeBSD is still there, do we really need the `--no-backup-if-mismatch` it's been months that we are asking to remove this option and use `--posix`.

Should we be really forced to fork this project ???

tmilker:
I think I determined the real fix for this issue that will make everyone happy(haven't tested on MacOS though):  composer-patches doesn't pass the -t or --batch command like it probably should.  I don't think anyone will ever want to manually type a filename(or have composer-patches stall, waiting for one during an automated deploy) and patch should just fail if it needs one.  A nice side effect is that on FreeBSD(tested on 11.3), if you specify -t it will ignore the invalid --no-backup-if-mismatch argument completely!

@cweagans Please modify composer-patches to add the -t command line argument to patch.  I'd love to remove my ugly, ugly hack of passing patch through my own patch script that adds it.

kentr:
For anyone still facing this, try `vaimo/composer-patches`.  It's working for me on FreeBSD w/ `patch 2.0-12u11`.

[...]

PR #196, opened 2018-02-09
Prevent interactive mode using the --batch flag

martijnhoutman:
We sometimes encounter patches that need to be applied on different directory levels, e.g. -p0, -p1, -p2, ...

The patch command however sometimes breaks on a certain level, unable to find the file and switching to interactive mode, asking the user to enter the correct path. This breaks our continuous deployment system.

The patch command has the -t | --batch flag in order to prevent the interactive mode.

I also need to apply the patch found in https://github.com/cweagans/composer-patches/issues/182 to be compatible with our FreeBSD systems.

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -423,7 +423,7 @@ class Patches implements PluginInterface, EventSubscriberInterface
                 // --no-backup-if-mismatch here is a hack that fixes some
                 // differences between how patch works on windows and unix.
                 if ($patched = $this->executeCommand(
-                    "patch %s --no-backup-if-mismatch -d %s < %s",
+                    "patch %s --posix --batch -d %s < %s",
                     $patch_level,
                     $install_path,
                     $filename
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -420,8 +420,8 @@ class Patches implements PluginInterface, EventSubscriberInterface
         // the 'patch' command.
         if (!$patched) {
             foreach ($patch_levels as $patch_level) {
-                // --no-backup-if-mismatch here is a hack that fixes some
-                // differences between how patch works on windows and unix.
+                // --posix is added to patch in order to make sure
+                // patch behaves the same way on windows, linux and unix.
                 if ($patched = $this->executeCommand(
                     "patch %s --posix --batch -d %s < %s",
                     $patch_level,


martijnhoutman:
Hmm, the --posix flag is actually not a replacement for --no-backup-if-mismatch flag, as it prevents creating new files from patches.

Will have to search for an alternative.

waleedq:
I was testing with patch utility and the -t flag, the -t flag as its documented in man-pages would "assume that patches are reversed if they look like they are." which from my understanding to the reversed patches might lead to some issues if the patch is already applied in the current version. please correct me if i am wrong about this assumption. So instead we can use the -f flag as mentioned in the 'patch' man-pages

>        -f  or  --force
>           Assume that the user knows exactly what he or she is doing, and do
>           not ask any questions.  Skip patches whose headers do not say
>           which file is to be patched; patch files even though they have the
>           wrong version for the Prereq: line in the patch; and assume that
>           patches are not reversed even if they look like they are.  This
>           option does not suppress commentary; use -s for that.

But what's better than using -t or -f is to test the patch before applying it using the --dry-run as proposed in this pull-request

https://github.com/cweagans/composer-patches/pull/202

This way we can use the -f flag in the dry-run test, if the test passes then we can apply the patch without -f|-t flags because we will be 100% sure that the patch will apply without any questions from 'patch'. 

Thank you.

zrhoffman:
> Hmm, the --posix flag is actually not a replacement for --no-backup-if-mismatch flag, as it prevents creating new files from patches.

New files can be created using `--posix`, you just need to provide that file as an argument, rather than interactively.

<details><summary>Here it is failing interactively:</summary>

```shell
username@computer:directory [0]$ <non-unified-patch patch --posix
can't find file to patch at input line 1
File to patch: example-file
example-file: No such file or directory
Skip this patch? [y]
Skipping patch.
1 out of 1 hunk ignored
```
</details>

<details><summary>Here it is succeeding when a filename is provided:</summary>

```shell
username@computer:directory [0]$ <non-unified-patch patch --posix example-file
patching file example-file
```
</details>

All of the unit and acceptance tests pass for me in Travis CI, not sure why an acceptance test failed before.

cweagans:
We've been doing this for a while, I think. #447 has more support for individual arguments passed to various patchers. If you'd like, take a look at `src/Patcher/*` to make sure things are captured properly.

PR #204, opened 2018-04-12
Add a fix for lightning profile (and other profiles with alternative docroots)

Techwolf12:
The lightning profile uses the docroot folder. Composer patches can't work with this (with Drupal core).

This adds support for the lightning profile and possibly more.

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -297,6 +297,18 @@ class Patches implements PluginInterface, EventSubscriberInterface
         $manager = $event->getComposer()->getInstallationManager();
         $install_path = $manager->getInstaller($package->getType())->getInstallPath($package);
 
+		if ($package_name == 'drupal/core' && $install_path == 'docroot/core') {
+	      $install_path = 'docroot';
+	      if ($this->io->isVerbose()) {
+	        $this->io->write(sprintf("'%s [id: %s]' installation path has been altered for patching.", $package->getName(), $package->getId()));
+	      }
+	    } else if ($package_name == 'drupal/core' && $install_path == 'html/core') {
+	      $install_path = 'html';
+	      if ($this->io->isVerbose()) {
+	        $this->io->write(sprintf("'%s [id: %s]' installation path has been altered for patching.", $package->getName(), $package->getId()));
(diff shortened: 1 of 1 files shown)


src/Patches.php
```diff
@@ -297,6 +297,26 @@ public function postInstall(PackageEvent $event)
...
+            $install_path = 'html';
+            if ($this->io->isVerbose()) {
+                $this->io->write(sprintf(
+                    "'%s [id: %s]' installation path has been altered for patching.",
+                    $package->getName(),
+                    $package->getId()
+                ));
+            }
+        }
+
```
neclimdul: Could this be a generalized string munge? So if something used `www/core` or something in the future this path altering would just work?

cweagans:
Adding special cases for packages isn't something I'm thrilled about doing. I'm not sure I understand why this is necessary though. If you're patching core, it should be a patch to stuff in the `core/` directory, so it should just work.

If you're at the Nashville sprints, I'm happy to chat in person.

Techwolf12:
I agree with @neclimdul that this should be generalized, it is just my quick fix pulled directly from our internal fork. I'll try to improve it when I have time. 

@cweagans If you have a lightning profile, everything will be in the ./docroot folder (./ relative to the directory with composer.json). Drupal core patches are often (I believe always following the standards) applied to the root dir, not the core dir. So it should run in ./docroot instead of ./docroot/core.

cweagans:
@Techwolf12 I think there's some misunderstanding about how things are working here. While your change might get the patch to apply, I don't think it's the right approach for this plugin or for your project.

The install directory for the drupal/core package in your install is docroot/core. That's the entire package. If you're applying a patch that just touches files in the docroot/core directory, everything will work fine. If you have a patch that changes files outside of the core directory, then you're not just patching the `drupal/core` package. The files other than the `docroot/core/` directory are not managed by composer, so if you want to change those files, you should just change them and commit the result.

It's definitely not a problem to use alternative docroots though. That doesn't have anything to do with this plugin. As long as the composer installer knows where the install path is, composer-patches can patch it.

PR #205, opened 2018-04-13
Daniel

danielmartinezd:
patch levels for Magento 2


danielmartinezd:
In order to apply patches in Magento 2 it's required this extra path

danielmartinezd:
Solves issue #206

PR #210, opened 2018-05-08
Handle patches-ignore when re-patching

### LINKED ISSUE #112: Ignoring patches from dependencies doesn't work

christopher-hopper:
## Proposed Changes

Prevent removing a package so it can be re-installed and re-patched just
because patches-ignore is being used. Ensure the patches-ignore are
accounted for when comparing patches with applied patches.

## Related Issues

Fixes #112 

## Notify

- @alexdesignworks 
- @brantwynn

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -98,10 +98,20 @@ class Patches implements PluginInterface, EventSubscriberInterface {
       $installationManager = $this->composer->getInstallationManager();
       $packages = $localRepository->getPackages();
 
+      $extra = $this->composer->getPackage()->getExtra();
+      $patches_ignore = isset($extra['patches-ignore']) ? $extra['patches-ignore'] : array();
+
       $tmp_patches = $this->grabPatches();
       foreach ($packages as $package) {
         $extra = $package->getExtra();
         if (isset($extra['patches'])) {
+          if (isset($patches_ignore[$package->getName()])) {
+            foreach ($patches_ignore[$package->getName()] as $package_name => $patches) {
+              if (isset($extra['patches'][$package_name])) {
+                $extra['patches'][$package_name] = array_diff($extra['patches'][$package_name], $patches);
+              }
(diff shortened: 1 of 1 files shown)


danepowell:
I haven't had time to test this, but the approach makes sense to me, seems like it would help with #112.

christopher-hopper:
If you're looking to test, these commands may help:

*Use modified package repo*
```sh
composer config repositories.composer-patches \
  git https://github.com/christopher-hopper/composer-patches.git
```

*Require patch-1 branch aliased as v1.6.5*
```sh
composer require "cweagans/composer-patches:dev-patch-1 as 1.6.5"
```

I'm using it now wherever we needed `'patches-ignore'` to remove a Drupal Lightning patch. It's now working as expected.

cweagans:
Thanks for this!

brantwynn:
Thank you so much for this, @christopher-hopper - it's been a pain in our sides for a long time.

brantwynn:
I am trying to use this via the latest dev (HEAD) of this project and I see this error once it gets to patching. 

```  - Applying patches for drupal/core
    https://www.drupal.org/files/issues/2752961-130.patch (Clear Twig caches on deploys)


  [ErrorException]
  Trying to get property 'url' of non-object
```
Has anyone else seen this? Happy to file another issue, but wanted to note that it came up when switching over to the new version.

cweagans:
You're probably using the HEAD of `master`, which you definitely don't want to be doing right now. Use 1.x instead.

brantwynn:
Ahh. Got it. 

Can confirm this is working for us to ignore some of the `acquia/lightning` patches during both `composer install` and `composer update` Ref: https://github.com/acquia/df/commit/a8fddc5684e921b43960fce82ce79d907de7d861

Thanks!

PR #212, opened 2018-05-18
Split out patch resolution into Composer Capabilities

cweagans:
This enables users to extend patch resolution in their own plugin. This would allow someone to essentially define their own logic for finding patches to apply to a package.

A lot of this code was ported out of the 2.x-prototype branch.

Remaining tasks:
* [ ] Get `checkPatches()` working again. Since we're now resolving patches before the first package is installed/updated, checkPatches() doesn't have any of that data. The obvious solution is to run patch resolution earlier, but that might be tricky.
* [x] Dependency patch resolution.
* [x] Dependency patch resolution opt-out flag. Alternatively (or maybe in addition), some way to disable specific patch resolvers.
* [x] Test coverage


cweagans:
The coveralls error is interesting. This PR increases coverage from ~7% to ~32% if I'm not mistaken.

src/Resolvers/PatchesFile.php
```diff
@@ -0,0 +1,98 @@
...
+     *   A list of patches.
+     * @throws \InvalidArgumentException
+     */
+    protected function readPatchesFile($patches_file)
+    {
+        $patches = file_get_contents($patches_file);
+        $patches = json_decode($patches, true);
+
+        // First, check for JSON syntax issues.
+        $error = json_last_error();
```
ryanaslett: Seems like you could use php's json_last_error_msg() and skip this whole switch/case statement below.

Issue #56, opened 2016-07-05
Re-running Composer Update fails because of Modified Files from previous patch


ndrake86:
After patching a module, unless you remove the directory where the patched module is located, when you re-run `Composer Update` you will get a message about "the package has modified files" which you have to discard or the module will fail to download and be re-patched. Below is an example: 

```
Gathering patches for root package.
Gathering patches for dependencies. This might take a minute.
  - Updating drupal/ajax_comments dev-8.x-1.x (44b6bf2 => 44b6bf2)
    The package has modified files:
    M ajax_comments.module
    M src/Utility.php
    Discard changes [y,n,v,d,s,?]? n
    Update failed (Update aborted)
    Would you like to try reinstalling the package instead [yes]? yes
  - Removing drupal/ajax_comments (dev-8.x-1.x)
    The package has modified files:
    M ajax_comments.module
    M src/Utility.php
    Discard changes [y,n,v,d,?]? y
  - Installing drupal/ajax_comments (dev-8.x-1.x 44b6bf2)
    Cloning 44b6bf2

  - Applying patches for drupal/ajax_comments
```

ndrake86:
Composer version: 1.1.3
Composer-patches: 1.5.0

ndrake86:
Sorry about the spam issue creation. No idea why github created 6 issues I closed them but if you want to delete those please do, I would have if I could.

ndrake86:
This also only looks to be an issue when the module you are pulling in is based on a dev branch (locked to a specific commit or not). If you are using a alpha, beta, etc. This does not seem to apply. Might have to do something like run a git clean or git reset --hard on that branch to clear out the patched changes.

cweagans:
I ran into this when writing the plugin originally. This is the reason for removing packages and repatching them whenever you run `composer install`. Can you please supply a complete `composer.json` that reproduces the problem?

ndrake86:
> Can you please supply a complete composer.json that reproduces the problem?
Sure, here you go: 
```
{
  "name": "composer-patch-issue",
  "description": "Error in patching Dev branches.",
  "type": "project",
  "minimum-stable": "dev",
  "prefer-stable": true,
  "repositories": [
    {
      "type": "composer",
      "url": "https://packagist.drupal-composer.org"
    }
  ],
  "require-dev": {
    "behat/mink-zombie-driver": "~1.2",
    "drupal/drupal-extension": "~3.0",
    "drush/drush": "~8.0",
[...]

PR #215, opened 2018-06-04
Fix applying patches to an existing patched dependency

deviantintegral:
This test shows the bug at #56 is still affecting master. This should fail with:

```
  - Applying patches for drupal/migrate_upgrade
    https://www.drupal.org/files/issues/2018-03-17/2915726-13.migrate_upgrade.Drush-9-port-of-commands.patch (Drush 9 support)
   Could not apply patch! Skipping. The error was: Cannot apply patch https://www.drupal.org/files/issues/2018-03-17/2915726-13.migrate_upgrade.Drush-9-port-of-commands.patch
```

TODOs:

- [ ] Improve how we handle fixtures with lock files, as currently they are removed in cleanup. I'm wondering if we can use `git clean` instead.
- [ ] Actually find and fix the bug
- [ ] See if we can simplify the test cases. I was having trouble reliably reproducing without these existing lock files. Hopefully once the bug is found and fixed a test case becomes more obvious.

### DIFF
--- a/src/PatchCollection.php
+++ b/src/PatchCollection.php
@@ -9,9 +9,17 @@ class PatchCollection implements \JsonSerializable
      *
      * Keys are package names. Values are arrays of Patch objects.
      *
-     * @var array
+     * @var \cweagans\Composer\Patch[]
      */
-    protected $patches;
+    protected $patches = [];
+
+  /**
+   * @return array
+   */
+    public function getPatches()
+    {
+        return $this->patches;
+    }
 
     /**
      * Add a patch to the collection.
--- a/src/PatchCollection.php
+++ b/src/PatchCollection.php
@@ -9,7 +9,7 @@ class PatchCollection implements \JsonSerializable
      *
      * Keys are package names. Values are arrays of Patch objects.
      *
-     * @var \cweagans\Composer\Patch[]
+     * @var array
      */
     protected $patches = [];
 
--- a/src/Plugin/Patches.php
+++ b/src/Plugin/Patches.php
@@ -217,25 +217,6 @@ class Patches implements PluginInterface, EventSubscriberInterface, Capable
             }
         }
 
-        $localRepository = $this->composer->getRepositoryManager()->getLocalRepository();
-        $packages = $localRepository->getPackages();
-        foreach ($this->patchCollection->getPatches() as $package_id => $patches) {
-            foreach ($packages as $package) {
-                if ($package->getName() == $package_id) {
-                    break;
-                }
-            }
-            $uninstallOperation = new UninstallOperation(
-                $package,
-                'Removing package so it can be re-installed and re-patched.'
-            );
-            $this->io->write('<info>Removing package ' .
-            $package_id .
-            ' so that it can be re-installed and re-patched.</info>');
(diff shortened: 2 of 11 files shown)


cweagans:
This behavior is _so_ strange. I need to figure out how to get a debugger attached to the plugin running in this test to be able to really understand what's going on. I tried reproducing this with something more similar to the existing test cases, but that didn't get me anywhere useful.

I did verify that you can collapse this into one test fixture though (composer.json.1, composer.lock.1, composer.json.2, and composer.lock.2, and then copy *.1 into place for the first composer run, then *.2 into place for the second), but that's not as important right now.

If you don't beat me to it, I'll set aside a weekend for this after some personal things settle out.

deviantintegral:
For debugging, I've been altering the test locally to run something like:

```
$I->runShellCommand('COMPOSER_DISCARD_CHANGES=1 COMPOSER_ALLOW_XDEBUG=1 XDEBUG_CONFIG="idekey=PHPSTORM" php -dxdebug.remote_enable=1 -dxdebug.remote_mode=req -dxdebug.remote_port=9000 -dxdebug.remote_host=127.0.0.1 -dxdebug.remote_connect_back=0 -d zend_extension=/usr/local/Cellar/php72-xdebug/2.6.0/xdebug.so composer.phar install -n');
```

But, I've found it easier to just run that from a normal terminal so I can control the vendor directories.

It looks like on master, the current issue is that existing packages that need to be re-patched are not removed:

```
//            ScriptEvents::PRE_INSTALL_CMD => array('checkPatches'),
//            ScriptEvents::PRE_UPDATE_CMD => array('checkPatches'),
```

I'm not seeing anywhere else in `resolvePatches` that duplicates the uninstall operation.

cweagans:
Yeah, that's intentional. `checkPatches` is probably/hopefully going away. I spent some time mapping out new workflows last night, and I think I might be able to avoid that entirely. I'm not 100% sure though. My hope is that 2.x can take a more nuanced approach than "¯\_(ツ)_/¯ nuke it from orbit, download it again, and patch it again" :)

deviantintegral:
Does it make sense to merge this in since it fixes a bug, or do you want to take a crack at another method first? If so, I'll hold off on cleaning this up.

cweagans:
Let's wait for now - this actually causes a different bug that people have seen in the past. Essentially, if you `composer update foo/bar`, only `foo/bar` should be deleted in the repatching section, but in this case, _all_ patched packages would be deleted and wouldn't necessarily be reinstalled.

That's part of the reason why I wanted to add a bit more structure around this process. Give me a couple days to get that stuff into place, and then we can pick at least the test case out of this one. I definitely don't want this to break again in the future, especially after you spent a bunch of time tracking down the problem.

deviantintegral:
> all patched packages would be deleted and wouldn't necessarily be reinstalled.

Yep, I actually just fixed that. I'm going to get this to a place where things aren't quite so broken for a project, and then I'll leave it there. Thanks!

cweagans:
I'll try to get stuff to a good state as quickly as possible. Thanks again for your work here - this is great :)

cweagans:
I think #447 should handle most of this, but please let me know if that's not the case.

PR #216, opened 2018-06-05
Document minimum git version

deviantintegral:
I was building on a server with Ubuntu 16.04 and it ships with git 1.9.1. In those cases, applying patches to non-git dependencies silently fails as no error message or exit code is returned by `git apply`.

This PR just updates the README; IMO it's probably too much complexity to parse `git --version` given the format of the string isn't stable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ Simple patches plugin for Composer. Applies a patch from a local or remote file 
 
 ## Support notes
 
+* Git >= 2.15.1 is required to properly apply patches.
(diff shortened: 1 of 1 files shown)


cweagans:
:+1: good call. I think we're going to start moving in the direction of using `patch` for everything in the near future (just for the sake of predictable behavior), so hopefully this won't be a concern for too much longer.

stucki:
This should not be necessary anymore after merging #227.

cweagans:
I've made a note to capture this and/or #227 in #447.

PR #217, opened 2018-06-11
Clarify how the `enable-patching` parameter actually works

pfrenssen:
The documentation about the `enable-patching` parameter is worded in a way that implies that it directly controls whether or not patches from dependencies will be applied.

For the 2.x branch we could revisit this and perhaps provide an option `apply-patches-from-dependencies`, but for now let's just update the documentation so it is clear how it works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Then your `composer.patches.json` should look like this:
 
 ## Allowing patches to be applied from dependencies
 
-If you want your project to accept patches from dependencies, you must have the following in your composer file:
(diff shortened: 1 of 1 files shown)


cweagans:
I think this one should probably be against the `1.x` branch -- the `2.x` branch has some different behaviors (that will be changing a bit more).

pfrenssen:
You're absolutely right! I created this through the GitHub web UI and didn't notice it was against the 2.x branch. I'll rectify.

pfrenssen:
@cweagans I have rebased this on the 1.x branch and force pushed. This didn't trigger the tests but it should be fine since it is just an update to the Readme.

PR #220, opened 2018-07-04
[1.x] Check that patches_applied has a count > 0 before uninstalling package

jameshalsall:
Fixes an issue where `composer-patches` would always remove a package and then re-install it in a Magento 2 project because the `patches_applied` was an empty array.


cweagans:
:+1: thanks!

PR #224, opened 2018-07-11
Array to string exception if multiple patches

mxr576:
Re-roll of #194


cweagans:
Thanks!

jibran:
@cweagans can we please have a new release for 1.x branch?

mike-potter:
Is there ANY way we can get a new release with this?  It's an old issue and still causing trouble these days.  Don't really like linking to composer-patches-dev.

Issue #172, opened 2017-11-22
Use patch instead of git apply


danepowell:
Following up on the conversation here: #165

Currently, composer-patches tries to use `git apply` to apply patches, and falls back to GNU patch if that fails. I think we should reverse this logic, or eliminate `git apply` altogether.

The reason being that you can't run `git apply` [inside of a git repo that's a parent of the project getting patched](https://stackoverflow.com/a/27283285). This probably describes 90% of the environments where composer patches is used, so `git apply` would almost never work.

What I mean by this: let's say you have a project _foo_ that depends on package _bar_. You obviously keep _foo_ in git, so your directory structure looks like this:

    foo/.git
    foo/composer.json
    foo/vendor/bar/bar

And now you want to apply a patch to bar. You can't do this with git apply, because `foo/` is the git repo root, but git apply expects `foo/vendor/bar/bar` to be the repo root.

@cweagans mentioned in https://github.com/cweagans/composer-patches/pull/165#issuecomment-343317202:
> git is the default because pretty much everyone has that installed. Fewer people have patch installed

I'd be surprised if that's true. I've never run across a *nix environment that didn't have patch installed. The only place I can imagine this being true is for Windows users running outside of the Ubuntu subsystem (which is not a recipe for success to begin with).

cweagans:
-1 for these reasons (in no particular order):

1.  This is in the readme, and it alleviates part of the issue:

```
"config": {
    "preferred-install": "source"
  },
```

2. `git apply` typically is more tolerant of fuzz (i.e. patch application is successful more often with `git apply` than `patch`).

3. The longer term goal here is to use the VCS-appropriate patcher (`svn apply`, `hg import`, etc), and then fall back to lesser methods if necessary (`patch`, a pure PHP patcher, etc).

4. `patch` is ubiquitous, but not consistent across environments. For instance, https://github.com/cweagans/composer-patches/issues/159 shows an issue where the actual command line flags for patch are different on FreeBSD. 

5. The linked stackoverflow comment describes a behavior from 2014. When I wrote the first version of this plugin, applying patches to nested git repos was the *only* thing it was used for. I'm inclined to think that things have changed a bit since then, but if not, this sounds like a good opportunity to engage with the `git` community to fix a bug.

ptmkenny:
I'd like to second the original request by @danepowell.  I just got bitten by the bug caused by the silent failure of git apply in #174.

[...]

PR #227, opened 2018-07-13
Fix #174: Compatibility with older Git versions

### LINKED ISSUE #174: Patches fail to apply with older versions of Git

stucki:
This PR applies the same changes as in #175 but for the current master branch.

It fixes compatibility of composer-patches with older Git versions, as reported in various issues:

- #172 
- #174 
- #216 should become unnecessary as well with this change

### DIFF
--- a/src/Plugin/Patches.php
+++ b/src/Plugin/Patches.php
@@ -385,8 +385,6 @@ class Patches implements PluginInterface, EventSubscriberInterface, Capable
             $downloader->copy($hostname, $patch_url, $filename, false);
         }
 
-        // Modified from drush6:make.project.inc
-        $patched = false;
         // The order here is intentional. p1 is most likely to apply with git apply.
         // p0 is next likely. p2 is extremely unlikely, but for some special cases,
         // it might be useful. p4 is useful for Magento 2 patches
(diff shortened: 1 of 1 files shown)


cweagans:
I've made a note to capture this in #447.

PR #229, opened 2018-07-20
Fix PHP 5.3 compatibility issue

### LINKED ISSUE #228: 1.6.5 is not compatible with PHP 5.3
Commit 49dbb8f618b6f3d25e67f127dd400d487a6afd67, which was merged and released in version 1.6.5, is not compatible with PHP 5.3:

> Parse error: syntax error, unexpected '[', expecting ')' in vendor/cweagans/composer-patches/src/Patches.php on line 381

However, the `composer.json` file for 1.6.5 says the release should be compatible with PHP 5.3.

I'd therefore recommend either fixing the syntax error or bumping the minimum PHP version.

colinodell:
Fixes #228


cweagans:
Thanks!

Issue #221, opened 2018-07-06
acts during 'composer update --lock', when it shouldn't


joachim-n:
AFAIK, 'composer update --lock' should only update the .lock file.

However, I get this output:

```
$ composer update --lock
    1/9:	https://packages.drupal.org/8/drupal/provider-2015-3$84d8c556a8f01336dc4264cc4557466a11c808d237ef9e8d427befa16f1af84d.json
    2/9:	https://packages.drupal.org/8/drupal/provider-2016-2$af04b1ed153ab3c869a90878768e59ae62532349f2ec751cabdc5379f17b32d3.json
    3/9:	https://packages.drupal.org/8/drupal/provider-2017-4$d2397a6675ba2bff38ce5f04f84cb22c62227ac6bf398476d68ba6e1a7232ae5.json
    4/9:	https://packages.drupal.org/8/drupal/provider-2017-2$a1651b5504506eb92f6e20b49e772c846d9d1ba4b6c802dc8f71b98e0671d2b1.json
    5/9:	https://packages.drupal.org/8/drupal/provider-2018-1$d7c3e6bcb1ac0d79727f314bdfb42d65649d5694d5f8fd329b802baae82adc75.json
    6/9:	https://packages.drupal.org/8/drupal/provider-2017-3$e16b48d5253d2279e6340f621ba7630a09f3c0af7aab0d25bfd924c37490beaf.json
    7/9:	https://packages.drupal.org/8/drupal/provider-2016-3$6adb2f66db28383a504de4ce9ee491cadaad6a7fe2d672807c51ca6d2c0080e7.json
    8/9:	https://packages.drupal.org/8/drupal/provider-2018-2$3921f8e9c4773aa201f556e778f1b4655a475b2ae17baf5ee10b3f4dc651f508.json
    9/9:	https://packages.drupal.org/8/drupal/provider-2017-1$6c34463523e459d87826703dbd23a7128e4552ceec799c273ff1a399f7b37197.json
    Finished: success: 9, skipped: 0, failure: 0, total: 9
    1/10:	http://packagist.org/p/provider-archived$a8b5b3db8189b9e9c26ef8fd8afe93ee7f97826f90d39dbfe802dd6a6c52ab49.json
    2/10:	http://packagist.org/p/provider-latest$37b92d24bbdc31bb31a8e9e1917a740e651d7ed491d64a8fa0d0abdd1ae4299e.json
[...]

PR #232, opened 2018-08-15
make sure not to apply/check patches when running with --lock

allanpaiste:
Make sure that patching is switched off when user only wants to update composer.lock hash by running composer update with --lock flag set.

Addresses https://github.com/cweagans/composer-patches/issues/221

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -415,6 +415,10 @@ class Patches implements PluginInterface, EventSubscriberInterface {
    *   Whether patching is enabled. Defaults to TRUE.
    */
   protected function isPatchingEnabled() {
+    if ($this->isLockHashUpdate()) {
+      return FALSE;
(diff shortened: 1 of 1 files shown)


weitzman:
Seems reasonable. I was wondering why we do that.

cweagans:
I think there's a property on one of the Composer objects that we can look at, isn't there? This seems like a really reasonable thing to do - just not 100% sure on the approach.

davereid-pfg:
Just a heads up that there's lots of documentation that when you change patches in your composer.json file, we've been always advised to run composer update --lock to get the new patches applied. There is no `composer update nothing` kind of command.

cweagans:
#447 should handle this. I've made a note to ensure that there's some mechanism to force patches to be applied.

PR #242, opened 2018-11-29
Allowing original files to be backed up.

zerustech:
Add a new extra boolean parameter `backup-before-applying-patches`, when
this parameter is `true`, the `patch -b` command will be used to backup
the original files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,18 @@ Composer [blocks](https://getcomposer.org/doc/06-config.md#secure-http) you from
 }
 ```
 
+## Allowing original files to be backed up
+
+If you want to backup the original files before applying patches, you must have the following in your composer file:
+
+```json
+{
+  "extra": {
+      "backup-before-applying-patches": true
+  }
+}
+```
+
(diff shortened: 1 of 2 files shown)


cweagans:
Why do you need this? Shouldn't the original contents be viewable in both the source repository for your dependency and in the actual patch file?

zerustech:
Thank you for your quick response @cweagans Consider the following scenario as an example:
1. We are working on a project that depends on a third party package `abc/foo`.
2. We want to maintain some local changes to `abc/foo` at this moment.
3. Then we create patch files that contain those changes.
4. We continue to develop the project, and some day we need change those files again.
5. If we have the `.orig` backup files, we can restore each file to its original version, otherwise, we will have to disable the patches from `composer.patches.json`, re-install the `abc/foo` package, and work on the changes.

cweagans:
Probably not going to do this, sorry. If you really want this, #447 gives you events that you can hook into to do whatever you'd like with the original files.

PR #243, opened 2018-12-09
Added composer & io to dispatched events, along with on patch fails event

waleedq:
-Added composer & io to dispatched events to make PatchEvent  more standard and to comply with Composer builtin events such as PackageEvent

- Added PATCH_APPLY_ERROR event that will be dispatched when the patch fails to apply


moerazem:
To better understand the purpose of this pull-request, here's a documentation of what we have implemented.

https://docs.varbase.vardot.com/updating-varbase/handling-patches-when-updating

Natshah:
Thank you very much for **cweagans/composer-patches**
My hope that this pull request got accepted and merged
Not sure what is messing to make it pass the automated testing
We had been testing this feature in many websites, which eases the work .
[Add a comment in composer.json to highlight why we're using the vardot/composer-patches repo Vardot/varbase-project#45](https://github.com/Vardot/varbase-project/issues/45)

Thank you for your time maintaining this very important composer plugin
my other hope that if Drupal core start to use it directly in core. or even one more smarter way too

PR #218, opened 2018-06-15
Correctly resolve relative local paths pointing to patch files from dependencies

rlmumford:
Update from PR #161

PR #247, opened 2019-01-24
#218 - Correctly resolve relative local paths pointing to patch files from dependencies - Follow up

drupol:
This is an updated version of PR #218.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 vendor/
+tests/_data/test-package/src/
--- a/src/Patch.php
+++ b/src/Patch.php
@@ -40,6 +40,13 @@ class Patch implements \JsonSerializable
     public $depth;
 
     /**
+     * The package that provides the patch.
+     *
+     * @var \Composer\Package\PackageInterface $provider
+     */
+    public $provider;
+
+    /**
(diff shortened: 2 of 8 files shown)


drupol:
I don't know why the coveralls is not triggered anymore but the coverage have increased since I added the tests.

rlmumford:
Is this just waiting on the test coverage?

yanniboi:
I would also love to see some progress on this. Anything we can do to push it forward?

drupol:
Yes I think increasing the test coverage would be a nice and the maintainer will be pleased.

rlmumford:
@drupol unfortunately I've never worked with coveralls before and I'm not sure what would be required to increase that coverage stat.

drupol:
You do not need coveralls, you just need to write more tests, coveralls will compute the coverage by itself.

iamweird:
This PR seems to be obsolete due to https://github.com/cweagans/composer-patches/pull/267.

louis-cuny:
I would be happy to see this PR merged, how can I help ?

cweagans:
Resolving patches from dependencies isn't a thing in 2.x, so neither is applying patches stored relative to the root of a dependency.

PR #248, opened 2019-01-25
Add missing ext-json in composer.json.

drupol:
We are using the PHP function `json_decode` in some places in the code.

Therefore, we need to ensure that users has the proper PHP extension "ext-json".

In order to do that, we need to modify the `composer.json` and add

`"ext-json": "*"`

in the `require` section.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,6 +14,7 @@
   ],
   "require": {
     "php": ">=5.6.0",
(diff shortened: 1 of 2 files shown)


drupol:

drupol:
Thanks !

PR #249, opened 2019-01-25
Use strict comparison.

drupol:
Replace `!=` and `==` with respectively `!==` and `===`.


drupol:

PR #250, opened 2019-01-25
Use single quote.

drupol:
Use single quote everywhere.

### DIFF
--- a/src/Plugin/Patches.php
+++ b/src/Plugin/Patches.php
@@ -378,7 +378,7 @@ class Patches implements PluginInterface, EventSubscriberInterface, Capable
             $filename = realpath($patch_url);
         } else {
             // Generate random (but not cryptographically so) filename.
-            $filename = uniqid(sys_get_temp_dir() . '/') . ".patch";
+            $filename = uniqid(sys_get_temp_dir() . '/') . '.patch';
 
             // Download file from remote filesystem to this location.
             $hostname = parse_url($patch_url, PHP_URL_HOST);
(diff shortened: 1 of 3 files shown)


drupol:

cweagans:
> Use single quote everywhere.

Why?

drupol:
We usually use single quote when using simple strings. For more complex strings with variables, we should use double quote.

See: http://php.net/language.types.string#language.types.string.syntax.single
See: https://phptherightway.com/pages/The-Basics.html

Using single quotes is therefore faster (_I agree this is micro-optimization_).

drupol:
I'm going to rebase this now.

cweagans:
I'd rather not do this, mostly because this style probably won't be maintained as development progresses.

drupol:
Oh sad... Let me try again.

We could enforce it through multiple ways:

# With PHP-CS-Fixer: The Symfony coding standard uses it everywhere through PHP-CS-Fixer, see: https://github.com/FriendsOfPHP/PHP-CS-Fixer (`CTRL+F single_quote`).
# With .editorconfig: A very simple file at the root of the project with a couple of configuration options, Drupal 8 is using it as well and there's a directive to configure the quote type (https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties). When using the .editorconfig plugin on your favorite IDE, it will configure it thanks to that file. It's very handy.
# Drupal is using it by default: https://www.drupal.org/docs/develop/standards/coding-standards#quotes

I think the lightest way to enforce it would be to use .editorconfig file.

PR #251, opened 2019-01-25
Update code style.

drupol:
Update general code style:
* Remove unneeded imports.
* Add missing properties
* Change conditional structures and use early returns (or early continue)
* Use `::class` whenever it's possible.
* Remove unneeded variables.
* Replace `call_user_func_array()` with something faster.

### DIFF
--- a/src/Plugin/Patches.php
+++ b/src/Plugin/Patches.php
@@ -21,10 +21,10 @@ use Composer\Plugin\Capable;
 use Composer\Plugin\PluginInterface;
 use Composer\Installer\PackageEvents;
 use Composer\Script\Event;
-use Composer\Script\ScriptEvents;
 use Composer\Installer\PackageEvent;
 use Composer\Util\ProcessExecutor;
 use Composer\Util\RemoteFilesystem;
+use cweagans\Composer\Capability\CoreResolverProvider;
 use cweagans\Composer\Capability\ResolverProvider;
 use cweagans\Composer\PatchCollection;
 use cweagans\Composer\Resolvers\ResolverBase;
@@ -75,6 +75,11 @@ class Patches implements PluginInterface, EventSubscriberInterface, Capable
      * @var PatchCollection $patchCollection
      */
     protected $patchCollection;
+
+    /**
+     * @var array
+     */
+    private $installedPatches;
 
     /**
      * Apply plugin modifications to composer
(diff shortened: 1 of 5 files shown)


drupol:

src/Resolvers/DependencyPatches.php
```diff
@@ -27,18 +25,24 @@ public function resolve(PatchCollection $collection, PackageEvent $event)
...
         $operations = $event->getOperations();
         foreach ($operations as $operation) {
-            if ($operation->getJobType() == 'install' || $operation->getJobType() == 'update') {
-                // @TODO handle exception.
-                $package = $this->getPackageFromOperation($operation);
-                /** @var PackageInterface $extra */
-                $extra = $package->getExtra();
-                if (isset($extra['patches'])) {
-                    $patches = $this->findPatchesInJson($extra['patches']);
-                    foreach ($patches as $package => $patch_list) {
```
cweagans: While we didn't use the `$package` var here, I think this style provided some clarity around the data structure. Can you revert back to this notation?

cweagans: https://www.facebook.com/groups/millennialshow/permalink/2074484859516440/

cweagans: (just the `foreach` line - the rest of the style changes are fine)

src/Plugin/Patches.php
```diff
@@ -76,6 +76,11 @@ class Patches implements PluginInterface, EventSubscriberInterface, Capable
      */
     protected $patchCollection;
 
+    /**
+     * @var array
+     */
+    private $installedPatches;
```
cweagans: Was this mistakenly added?

drupol: Nope, `$this->installedPatches` is used on line 97 of the same file :-)

src/Resolvers/DependencyPatches.php
drupol: I updated it, changed the variable into something more meaningful without wasting memory with unused vars.

drupol:
@cweagans Good for you ?

cweagans:
Most of this is obsolete. See #447 for latest.

drupol:
Thanks for being so responsive!

PR #265, opened 2019-04-09
Use empty() instead of isset() to check for empty arrays in RootComposer (test coverage 100%)

ElijahLynn:
This gets RootComposer to full coverage.


ElijahLynn:
https://www.php.net/manual/en/types.comparisons.php

Issue #39, opened 2016-04-12
File based patches aren't resolved in dependencies


deviantintegral:
I have a patch like the following:

```
            "drupal/core": {
                "Add option to disable settings.php hardening": "patches/drupal/1232572.43-sites-dir-hardening.patch"
            },
```

If that patches line is in an dependency, composer-patches looks for the file in the patches directory of the root package, instead of in the vendor directory of the package that adds the patch.

URL patches work around this nicely, but aren't great in scenarios where a patch is private or isn't accepted upstream.

cweagans:
How critical is it that the patch is stored in `vendor/drupal/core`? I'm not sure why you'd want it there, because a `composer update` would nuke that entire directory. In addition, `composer install` might actually wipe it out too, since patched dependencies are deleted, re-installed, and repatched if the patch definition has changed.

littlethoughts:
Hi @cweagans, I am also interested in using file based patches. Of course I wouldn't store them in a directory that contains generated files. The use case is that I have custom patches which I would like to apply, they aren't uploaded anywhere. Is this something that could be added?

deviantintegral:
The patch itself is stored in the root of the project in a `patches` directory, not in `vendor/...`.

cweagans:
It took some time, but I see what you're saying now :)

If package A depends on package B, and package B defines a local path to a patch, the path should be relative to the installation root of package B. Makes sense, but might be somewhat problematic to implement.

Johnzzon:
This is a feature we'd benefit of in our latest project. We're building a profile for use in several of our sites, and this profile needs some core patches (which can be very use case specific changes, that we don't want to upload to an issue on drupal.org). We can host the patches ourselves in the meantime, but ideally we'd want to bundle them with the profile and not be dependent on external hosting.

Ambient-Impact:
Question: I've run into a very similar problem while using [wikimedia/composer-merge-plugin](https://github.com/wikimedia/composer-merge-plugin), in that patching fails if I specify a relative patch path in a Drupal module I'm developing when that's merged into the root `composer.json`, likely because the merge plug-in doesn't know to resolve the path from being relative to the module to being relative to the root `composer.json`. Is there a way I can rewrite the paths somehow as a workaround?

cweagans:
`main` no longer resolves patches from dependencies like 1.x does. Probably not going to fix this, sorry.

Ambient-Impact:
[...]

PR #267, opened 2019-04-09
Remove dependency patch resolution

cweagans:
This is something that I've wanted to do for a while, but we haven't had the infrastructure to let people add it on to the plugin if they really need this functionality. I was hesitant to add it in the first place, because allowing some arbitrary package to patch another has some security risks associated with it (imagine some contrib module just nuking the entire filter module in drupal/core).

Since PatchResolvers are now Composer Capabilities, this plugin can be extended by another plugin. I'm happy to help someone build out their own plugin to provide this functionality if there's strong enough interest. There is also some movement toward making patch _application_ a Capability, so a third party plugin will be able to handle transforming the path to the patch file appropriately (to address issues like https://github.com/cweagans/composer-patches/issues/39).

From a maintenance burden standpoint, this functionality has probably been one of the more difficult parts of the plugin. Aside from the security risks, it's been pretty error prone, and there are a ton of instances where multiple modules require the same patch (which causes problems - can't apply the same patch twice, the internal data structure doesn't really tolerate duplicate patches that well, etc).


tests/_support/_generated/UnitTesterActions.php
```diff
@@ -540,9 +565,45 @@ public function fail($message) {
      *
      * @param $exception string or \Exception
      * @param $callback
+     *
+     * @deprecated Use expectThrowable instead
```
ElijahLynn: Is this valid?

ElijahLynn:
:+1:

tests/_support/_generated/UnitTesterActions.php
cweagans: Yeah, it's generated code from codeception.

AndreasA:
@cweagans Any news when this will be released with an official version because I think it is rather important 😄

xurizaemon:
@AndreasA I had to look twice to realise that this PR **removes support** for resolving the path to patches in dependencies. From what I believe I see in current composer:

- Patches mentioned in dependences will be applied (if [enable-patching is set true](https://github.com/cweagans/composer-patches#allowing-patches-to-be-applied-from-dependencies), if you reference patches by network URLs (relative paths don't apply)
- Patches mentioned in dependencies won't have their paths resolved, so they will be looked for relative to the project root not the dependency root.
- An [external patch file reference](https://github.com/cweagans/composer-patches#using-an-external-patch-file) also will be looked for relative to the project root, not the dependency root.

This PR would have been released in [1.6.6](https://github.com/cweagans/composer-patches/releases/tag/1.6.6).

AndreasA:
@xurizaemon from what I can see the commit was merged forever ago but not yet included in a released version of the package which is weird. Personally, I think it makes more sense to specify all patches in the root only, to avoid weird side-effects and confusion.

AndreasA:
@xurizaemon I have now checked the code for v1.x and according to this code line https://github.com/cweagans/composer-patches/blob/1.7.2/src/Patches.php#L436 
the dependency patching is always enabled as soon as the root composer.json uses patches. 
At least prior to this change which is part of 2.x and still not released.

cweagans:
This pr only affects the 2.x release, which has no specific release date nor timeline. This will not be added to 1.x.

PR #282, opened 2019-08-29
Retry failed downloads once

### LINKED ISSUE #279: Retry failed download
For a while now, d.o has some infrastructure problems that causes failed downloads, the result looks like this:

Could not apply patch! Skipping. The error was: Content-Length mismatch, received 0 bytes out of the expected 3784

This is completely random and causes failed CI builds, deployments and so on.

Until that's dealt with, it would be nice of composer-patches could maybe just retry a download 1-2 times in case of such a generic error?

Berdir:
Fixes #279


cweagans:
This approach works for me. I think this can also go into the 1.x branch if you want it in a more immediate release.

PR #283, opened 2019-08-29
Retry failed downloads once

Berdir:
Same MR as #282, against the 1.x branch.

I think a release soon would be very helpful, the network issues seem to be more frequent. Every other build/deployment today seems to fail.


cweagans:
Agreed. 1.6.7 is now tagged and available on Packagist.

PR #296, opened 2020-01-08
Fix #294 - Make "git apply" working in non-root-git dirs

### LINKED ISSUE #294: Make git apply working in non-root-git directories
`git` is the right choise for appliing patches as it is available more than every other solution at the most dev/prod systems, espetially non-linux ones like Windows, where the `patch` alternative is not often installed (or at least it is not available from the `PATH`).

I have analysed the issue related with issue #148 ( https://stackoverflow.com/questions/24821431/git-apply-patch-fails-silently-no-errors-but-nothing-happens/#27283285 ):

Problem:
patch is not applied via `git apply` if outside `.git` directory, in this case `Skipped patch '<file path>'.` verbose warning is show.

Analysis:
1. trigerred here: https://github.com/git/git/blob/b02fd2accad4d48078671adf38fe5b5976d77304/apply.c#L4696
2. based on condition https://github.com/git/git/blob/b02fd2accad4d48078671adf38fe5b5976d77304/apply.c#L4681
3. the condition is not matched in case of not matching `state->prefix` (.git directory root path) https://github.com/git/git/blob/b02fd2accad4d48078671adf38fe5b5976d77304/apply.c#L2077
4. `state->prefix` is set solely here https://github.com/git/git/blob/b02fd2accad4d48078671adf38fe5b5976d77304/apply.c#L93
5. from `setup_git_directory_gently()` https://github.com/git/git/blob/99c33bed562b41de6ce9bd3fd561303d39645048/git.c#L416 line as the `git apply` command is marked with `RUN_SETUP_GENTLY` option
6. `setup_git_directory_gently()` function tries to find `.git` directory and then it tests it with `is_git_directory` https://github.com/git/git/blob/bc12974a897308fd3254cf0cc90319078fe45eea/setup.c#L311 , this function tests the presence of these files:
   - `.git/objects/`
   - `.git/refs/`
   - `.git/HEAD`

These files needs to be presented in order to make `git patch` working.

PR submitted.

mvorisek:
Fixes #294

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -525,34 +525,65 @@ class Patches implements PluginInterface, EventSubscriberInterface {
    *   TRUE if patch was applied, FALSE otherwise.
    */
   protected function applyPatchWithGit($install_path, $patch_levels, $filename) {
-    // Do not use git apply unless the install path is itself a git repo
+    // "git apply" works only if the install path is itself a git repo
+    // @see https://github.com/cweagans/composer-patches/issues/148
     // @see https://stackoverflow.com/a/27283285
-    if (!is_dir($install_path . '/.git')) {
-      return FALSE;
-    }
+    $filesToUnlink = [];
+    try {
+      if (!is_dir($install_path . '/.git')) {
+        // Simulate "git init" to fix the "git apply" limitation.
+        // @see https://github.com/cweagans/composer-patches/issues/294
+        if (file_exists($install_path . '/.git') || !mkdir($install_path . '/.git')) {
+          return false;
+        }
+        $filesToUnlink[] = $install_path . '/.git';
 
-    $patched = FALSE;
-    foreach ($patch_levels as $patch_level) {
-      if ($this->io->isVerbose()) {
-        $comment = 'Testing ability to patch with git apply.';
-        $comment .= ' This command may produce errors that can be safely ignored.';
-        $this->io->write('<comment>' . $comment . '</comment>');
+        if (!mkdir($install_path . '/.git/objects')) {
+          return false;
+        }
+        $filesToUnlink[] = $install_path . '/.git/objects';
+
+        if (!mkdir($install_path . '/.git/refs')) {
+          return false;
+        }
+        $filesToUnlink[] = $install_path . '/.git/refs';
+
+        if (!file_put_contents($install_path . '/.git/HEAD', "ref: refs/heads/master\n")) {
+          return false;
+        }
+        $filesToUnlink[] = $install_path . '/.git/HEAD';
       }
-      $checked = $this->executeCommand('git -C %s apply --check -v %s %s', $install_path, $patch_level, $filename);
-      $output = $this->executor->getErrorOutput();
-      if (substr($output, 0, 7) == 'Skipped') {
-        // Git will indicate success but silently skip patches in some scenarios.
-        //
-        // @see https://github.com/cweagans/composer-patches/pull/165
-        $checked = FALSE;
+
+      $patched = FALSE;
+      foreach ($patch_levels as $patch_level) {
+        if ($this->io->isVerbose()) {
+          $comment = 'Testing ability to patch with git apply.';
+          $comment .= ' This command may produce errors that can be safely ignored.';
+          $this->io->write('<comment>' . $comment . '</comment>');
+        }
+        $checked = $this->executeCommand('git -C %s apply --check -v %s %s', $install_path, $patch_level, $filename);
+        $output = $this->executor->getErrorOutput();
+        if (substr($output, 0, 7) == 'Skipped') {
+          // Git will indicate success but silently skip patches in some scenarios.
+          // @see https://github.com/cweagans/composer-patches/pull/165
+          $checked = FALSE;
+        }
+        if ($checked) {
+          // Apply the first successful style.
+          $patched = $this->executeCommand('git -C %s apply %s %s', $install_path, $patch_level, $filename);
+          break;
+        }
       }
-      if ($checked) {
-        // Apply the first successful style.
-        $patched = $this->executeCommand('git -C %s apply %s %s', $install_path, $patch_level, $filename);
-        break;
-      }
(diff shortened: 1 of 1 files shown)


mvorisek:
Hi @cweagans, can we merge this?

cweagans:
I appreciate the work and research that you've put into this, but I'm probably not going to merge this. If it's not a git directory in the first place, there's not a good reason to use Git to apply patches to that directory. `patch` should be more than adequate. I'm not willing to complicate this plugin to make it do something it really doesn't need to do - it's already fragile enough.

mvorisek:
@cweagans This fix is very important - it solves one very strong usecase - when `patch` (for ex. any standard WIndows) is not avaialble. Can you reconsider this? Thanks in advance!

samdark:
@mvorisek can't help in this since I don't get the use case.

mvorisek:
> 
> 
> @mvorisek can't help in this since I don't get the use case.

@samdark Any system, where `patch` is not within `PATH`, but `git` is. On linux, `patch` is typically available, but on Windows it is not. Patchin with `git` is already supported, but if the main package does not use `.git`, the `git patch` command will fail as it requires to be run in root git directory. This PR solves this issue by faking the root directory and `git patch` then works like `patch`.

samdark:
Ah, so it is backwards compatible platform compatibility fix?

mvorisek:
It is forward and backward compatible. I spent quite a lot of time to address this issue properly.

In past the `git patch` issue (ie. requirement of the dir `.git` dir to be present) was addressed by fallback to `patch` command - see https://github.com/mvorisek/composer-patches/blame/2e6f72a2ad8d59cd7e2b729f218bf42adb14f590/src/Patches.php#L395-L396 . But this approach is more hotfix than a real solution as `patch` is typically not available on non-unix systems. Several issues for this were already opened.

With this PR, the `git patch` issue is fully addressed.

samdark:
I see. Well, it makes sense as a fallback but it's @cweagans to decide.

cweagans:
Why not just install ‘patch’? https://chocolatey.org/packages/patch

There’s only so far we can go to work around missing host dependencies, and this seems like overstepping that for a very rare thing.

In 2.x, patch application will be a capability, so you can extend it in a separate plugin. I’m still not convinced that this needs to happen at all though.

mvorisek:
> Why not just install ‘patch’? https://chocolatey.org/packages/patch

this tools should work in any enviroment where composer does, right?

Currently, any users that:
a) use Windows,
b) do not have `patch` in `PATH`,
b) and do not have `.git` dir in composer root
are affected.

This affected group is large, it was already reported here many times and we can fix this issue very robustly with a few lines of code without any side effect.

In code, the current behaviour is described as:
> // In some rare cases, git will fail to apply a patch, fallback to using
> // the 'patch' command.

very likely because of this issue.

I hope I explained the current situation well.

Now, back to your question - if I distribute my app patched using this package, is convincing users to install any executable really a better solution than a few lines of fixes in plugin that can work of the box?

@cweagans, please, review this PR. It would really help me and I hope others as well. Thank you.

cweagans:
> this tools should work in any enviroment where composer does, right?

Yes, as long as the system dependencies are satisfied. You didn't answer my question though. Why can't/won't you install `patch` for Windows?

> without any side effect.

We don't have any way of verifying this assertion in 1.x. I've made this assumption before and ended up causing myself a lot of work to get the plugin working again quickly because I broke a workflow for somebody. I'm not thrilled about the idea of making this assumption for something that can be easily worked around by installing a common tool on your workstation.

> Now, back to your question - if I distribute my app patched using this package, is convincing users to install any executable really a better solution than a few lines of fixes in plugin that can work of the box?

I don't think it's unreasonable to expect that a developer has (or can have) both `git` and `patch` installed. This plugin has been happily operating under that assumption since 2014.

Again, what you want to do will be possible by extending this plugin in 2.x, but I just haven't had time to get it done.

samdark:
@cweagans you have to know that you need to install it. As far as I understand the issue, it currently fails silently without giving any clue.

cweagans:
Okay. _That_ is something that I could see fixing in 1.x.

mvorisek:
> Okay. _That_ is something that I could see fixing in 1.x.

@cweagans Can you please merge then?

I hit this issue again and the patch is highly welcomed also for linux - Alpine has no `patch` command installed by default, thus support with `git` is highly appreciated.

andypost:
@mvorisek in alpinelinux you need anyway to run `apk add patch` or install git because busybox's one is unable to apply binary patches

cweagans:
@mvorisek My apologies. I may not have communicated clearly enough. I'm willing to merge a PR that gives an explicit error message about `patch` not being installed. I'm not interested in merging this particular PR.

andypost:
Moreover it's [just 100KB](https://pkgs.alpinelinux.org/package/edge/main/x86_64/patch)   vs 20MB of git

mvorisek:
@cweagans thanks for info - this PR solves however the `git apply` issue completely with very low LOC it fixes support for two major usecases:

- `patch` not available, very common on Windows
- `patch` not avaiable, very common on lite linux distros like Alpine

I am not sure if I explained the issues well, but this PR:

- does not drop support of `patch` in any way
- it does not even add `git` patch support. It only fixes it, so patch can be applied by `git` when the project is not withing a git repo

Please reconsider this fix once again.

cweagans:
@mvorisek I did take the time to thoroughly understand the PR. The problem is not a lack of understanding. It's that I disagree with the premise. `patch` is _available_ for every platform that I can think of. The fact that it's _not installed_ isn't a problem that this plugin needs to solve.

In 2.x (if I can ever get it finished and out the door), you'll be able to write a separate plugin that extends this one. Via Composer Capabilities, you'll be able to define whatever method you like for applying patches (including creating a fake .git dir and using git apply if that's what you want to do for whatever reason).

In either case though, I'm really not willing to bring this into the core plugin. The solution is to simply install `patch` and call it a day.

PR #304, opened 2020-03-13
Use drupal/core instead of drupal/drupal in examples

### LINKED ISSUE #303: Update usage example with drupal/core instead of drupal/drupal
I'm guessing the existing example was written when the composer template for Drupal projects was still the de-facto way to set up Drupal composer projects.

I copied and pasted that example today for my own migration (https://github.com/geerlingguy/jeffgeerling-com/issues/4), and when I did that, nothing happened. I thought I was going crazy, but then realized there is no `drupal/drupal` package if you set up your Drupal project using the new Drupal.org official composer project setup... instead, it's `drupal/core`.

I'll submit a fix in a PR. Not a biggie, but it did trip me up and led me down a rabbit hole that led me to believe maybe this project was broken or my brain was broken. Nope! Just a documentation example that needs updating...

geerlingguy:
Closes #303.﻿

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,14 +24,14 @@ Example composer.json:
 {
   "require": {
     "cweagans/composer-patches": "~1.0",
-    "drupal/drupal": "~8.2"
+    "drupal/core-recommended": "^8.8",
   },
   "config": {
     "preferred-install": "source"
   },
   "extra": {
     "patches": {
-      "drupal/drupal": {
+      "drupal/core": {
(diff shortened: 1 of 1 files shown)


cweagans:
A [drupal/drupal package](https://packagist.org/packages/drupal/drupal) does actually exist, although it's probably not the right thing to be using in examples, so this sounds like a great change. Thanks!

You may also be interested in using the 1.x version of this plugin, since master is the development branch for 2.0.0. Not a great situation, but I thought 2.x was going to be available sooner when I forked 1.x off onto it's own branch. 🤷‍♂

PR #306, opened 2020-04-01
Add missing config in ignore patches example

jdreesen:
If I understand the docs correctly, `enable-patching` must be `true` because only then patches from dependencies would be installed.


cweagans:
You only need the `enable-patching` param if you do not have any patches defined in your root composer.json. It's a way to explicitly tell the plugin to look for patches in dependencies. Is this broken currently?

jdreesen:
No, apparently I misunderstood the docs. Thx for the clarification.

PR #309, opened 2020-04-14
Add support for Composer version 2

Ayesh:
Hi Cameron,
Composer 2 is about to be released, and it has `composer-plugin-api` version 2. This PR updates the version constraint in `composer.json` to `^1 || ^2` so we can support both composer versions.

See [What's new in Composer 2](https://php.watch/articles/composer-2) and [UPGRADE-2.0](https://github.com/composer/composer/blob/master/UPGRADE-2.0.md#for-integrators-and-plugin-authors) for more changes in API. Empty methods \cweagans\Composer\Plugin\Patches::deactivate() and \cweagans\Composer\Plugin\Patches::uninstall are added to make it compatible both versions.

I will make a PR to the `cweagans/composer-configurable-plugin` to update its `composer/composer` version constraint because this package depends on it. 

Related: composer/composer#8726

Thank you 🙏.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
   "require": {
     "php": ">=5.6.0",
     "ext-json": "*",
-    "composer-plugin-api": "^1.0",
+    "composer-plugin-api": "^1.0 || ^2.0",
     "cweagans/composer-configurable-plugin": "^1.0"
   },
   "require-dev": {
--- a/src/Plugin/Patches.php
+++ b/src/Plugin/Patches.php
@@ -511,4 +511,18 @@ class Patches implements PluginInterface, EventSubscriberInterface, Capable
         }
         return ($this->executor->execute($command, $output) === 0);
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function deactivate(Composer $composer, IOInterface $io)
+    {
+    }
+
+    /**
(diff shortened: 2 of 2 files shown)


cweagans:
Thanks for this! I'm going to hold off on merging this for the time being, since the plugin won't actually work in Composer 2 with this PR. If you don't mind, I'll build off of this once I have some time and then get it merged in!

Ayesh:
Thanks @cweagans for your time. May I ask what is broken with Composer 2? It took me some fiddling, but I could get the plugin to work with composer 2. 

I didn't make extensive tests with exclusion patterns and separate patch files, but my package was indeed patched when downloaded. 

Because this plugin depends on another library that brings the configuration trait, I had to add these two got repos as path repos in the test project composer.json, so they are loaded from my forks instead of the upstream. 

If there's anything I can help, I'd be eager to work on them. But of course please take time. Thank you.

CDRO:
This plugin is such a corner stone in most of our projects, that I'd be glad to sponsor some development time if it allows you to release a version that is compatible with composer v2

cweagans:
@CDRO  there is already a release. 1.7.0

CDRO:
Ah yes, thanks for pointing that out, sometimes I should wait for the second coffee to kick in before asking a question :-)

PR #311, opened 2020-04-14
Support busybox

stefanisak:
busybox is a sized optimized toolbox of common unix utilities including a version of patch for alpine.
This version doesn't support all parameters so applying patches will fail. This PR introduces another attempt to patch files with params supported by busybox version of patch.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,10 +15,10 @@
   ],
   "require": {
     "php": ">=5.3.0",
-    "composer-plugin-api": "^1.0"
+    "composer-plugin-api": "^1.0 || ^2.0"
   },
   "require-dev": {
-    "composer/composer": "~1.0",
+    "composer/composer": "^1.0 || ^2.0",
     "phpunit/phpunit": "~4.6"
   },
   "autoload": {
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -404,6 +404,15 @@ class Patches implements PluginInterface, EventSubscriberInterface {
       }
     }
 
+    // support busybox patch version
+    if (!$patched) {
+      foreach ($patch_levels as $patch_level) {
+        if ($patched = $this->executeCommand("cd %s && patch %s < %s", $install_path, $patch_level, $filename)) {
+          break;
+        }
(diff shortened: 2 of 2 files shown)


andypost:
Looks interesting, just not clear which restrictions (except no support for binary patches) should be noted in readme

cweagans:
I'm not excited about supporting weird versions of patch (BSD also has one of these that's been a persistent problem). Up to this point, I've been supporting GNU Patch (which is packaged for alpine -- `apk add patch`). Is there some way to identify this version of `patch` vs some other version of `patch`? I'm thinking that 2.x is going to have extensible patch application, so if there's some way to determine what kind of `patch` is installed, this would be possible.

Specific to this patch, I imagine that at least _some_ flag will need to be passed to `patch`. I know at least GNU Patch makes a mess if a patch doesn't apply correctly, sometimes creates backup files, etc.

alappe:
```
/var/www/html $ patch -v
patch: unrecognized option: v
BusyBox v1.30.1 (2019-06-12 17:51:55 UTC) multi-call binary.

Usage: patch [OPTIONS] [ORIGFILE [PATCHFILE]]

        -p N    Strip N leading components from file names
        -i DIFF Read DIFF instead of stdin
        -R      Reverse patch
        -N      Ignore already applied patches
        -E      Remove output files if they become empty
        --dry-run       Don't actually change files
/var/www/html $ echo $?
1
/var/www/html $
```

That would be one way, but the non-success status code needs to be handled.

```
/var/www/html $ ls -la `which patch`
lrwxrwxrwx    1 root     root            12 Oct 21  2019 /usr/bin/patch -> /bin/busybox
/var/www/html $
```

That would be another.

cweagans:
Probably won't do this in 1.x, but in 2.x (and in particular in the 2.x-wip branch that #447 is using), individual versions of patch can be supported. Can you take a look at `src/Patcher/*` in that PR and let me know if the `BsdPatchPatcher` does what you need? I'm open to including a busybox patcher there if needed.

PR #321, opened 2020-07-15
Add support for Composer version 2

hussainweb:
Cherry picking from #309

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1,47 +1,116 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
-        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "hash": "a2cd6826c202b7ebefe3050efc2a7b7f",
-    "content-hash": "0ed9361502c0f5f22a5b440c16640193",
+    "content-hash": "4a5c841252204815536a37cad51d347b",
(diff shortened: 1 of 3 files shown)


composer.lock
```diff
@@ -1,60 +1,129 @@
...
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "hash": "a2cd6826c202b7ebefe3050efc2a7b7f",
-    "content-hash": "0ed9361502c0f5f22a5b440c16640193",
+    "content-hash": "4a5c841252204815536a37cad51d347b",
     "packages": [],
     "packages-dev": [
+        {
+            "name": "composer/ca-bundle",
```
prudloff-insite, line 11: This looks like a lot of unrelated changes got caught in the cherry-pick.
I think you should revert the `composer.lock` file to 972fac4d2535b1bb201e0f20a84a2ce62555c3a1 then run `composer update --lock`.

hussainweb: I was wondering about that. Is there a reason there is a composer.lock. It typically isn't in version control for libraries. It doesn't get used anyway when requiring a package.

hussainweb: And these changes might be because I also changed the version constraint for `composer/composer` to `"^1.0 || ^2.0"`. That has introduced new packages.

prudloff-insite: Oh right, I didn't see you changed the composer/composer dev requirement.
That would indeed explain it.

mxr576:
I am doing my own testing and bumping issues, like this one: https://github.com/composer/composer/issues/9079

PR #327, opened 2020-08-18
Fix Travis config linter warnings

kdambekalns:
Travis said:

- root: deprecated key sudo (The key `sudo` has no effect anymore.)
- root: missing os, using the default linux
- root: key matrix is an alias for jobs, using jobs

This change fixes those warnings


cweagans:
Thanks!

PR #330, opened 2020-09-11
Replace getJobType with an instanceof check

hussainweb:
Composer 2 has subsequently changed the method name of `getJobType` to `getOperationType` in https://github.com/composer/composer/pull/8537.

The result is I get the error below when trying to run any composer operation that results in a patching operation.
```
PHP Fatal error:  Uncaught Error: Call to undefined method Composer\DependencyResolver\Operation\InstallOperation::getJobType() in /.../vendor/cweagans/composer-patches/src/Patches.php:179
Stack trace:
#0 [internal function]: cweagans\Composer\Patches->gatherPatches(Object(Composer\Installer\PackageEvent))
#1 phar:///.../bin/composer/src/Composer/EventDispatcher/EventDispatcher.php(165): call_user_func(Array, Object(Composer\Installer\PackageEvent))
#2 phar:///.../bin/composer/src/Composer/EventDispatcher/EventDispatcher.php(118): Composer\EventDispatcher\EventDispatcher->doDispatch(Object(Composer\Installer\PackageEvent))
#3 phar:///.../bin/composer/src/Composer/Installer/InstallationManager.php(368): Composer\EventDispatcher\EventDispatcher->dispatchPackageEvent('pre-package-ins...', true, Object(Composer\Repository\InstalledFilesystemRepository), Array, Object(Composer\DependencyResolver\Operation\InstallOperation in /.../vendor/cweagans/composer-patches/src/Patches.php on line 179
```

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -176,7 +176,7 @@ class Patches implements PluginInterface, EventSubscriberInterface {
     $operations = $event->getOperations();
     $this->io->write('<info>Gathering patches for dependencies. This might take a minute.</info>');
     foreach ($operations as $operation) {
-      if ($operation->getJobType() == 'install' || $operation->getJobType() == 'update') {
+      if ($operation->getOperationType() == 'install' || $operation->getOperationType() == 'update') {
         $package = $this->getPackageFromOperation($operation);
         $extra = $package->getExtra();
         if (isset($extra['patches'])) {
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -176,7 +176,7 @@ class Patches implements PluginInterface, EventSubscriberInterface {
     $operations = $event->getOperations();
     $this->io->write('<info>Gathering patches for dependencies. This might take a minute.</info>');
     foreach ($operations as $operation) {
-      if ($operation->getOperationType() == 'install' || $operation->getOperationType() == 'update') {
+      if ($operation instanceof InstallOperation || $operation instanceof UpdateOperation) {
         $package = $this->getPackageFromOperation($operation);
         $extra = $package->getExtra();
         if (isset($extra['patches'])) {


danepowell (approved):

src/Patches.php
```diff
@@ -176,7 +176,7 @@ public function gatherPatches(PackageEvent $event) {
     $operations = $event->getOperations();
     $this->io->write('<info>Gathering patches for dependencies. This might take a minute.</info>');
     foreach ($operations as $operation) {
-      if ($operation->getJobType() == 'install' || $operation->getJobType() == 'update') {
+      if ($operation instanceof InstallOperation || $operation instanceof UpdateOperation) {
```
staabm, line 179: This line should be fixed, so it will work for composer v1 and v2

hussainweb: Fixed how? I tested this (changed) line on both composer 1 and 2 and it works fine.

staabm: Ok cool, didnt expect that. Even better.

I will also test this patch tomorrow, in our pipeline <3

clxmstaab:
just confirmed: with this fix composer-patches works for us on composer v2

clxmstaab:
@cweagans would be great you could merge this patch, so we can more easily do beta testing with composer v2

clxmstaab (approved):

src/Patches.php
mxr576: Just to be sure, are we sure that it works with _both_ versions? (Without test coverage it is hard to tell ;S)
I remember that I created a different fix when I experimented with Composer 2x on my project.
https://github.com/cweagans/composer-patches/compare/1.x...mxr576:1.x#diff-cf7534e34ca749ac0a88fd7f5dbe0869

rodrigoaguilera: Test coverage is desirable but I don't think we should block this change waiting for it.

At first I also wrote a solution using `\Composer\Composer::RUNTIME_API_VERSION` but this solution is cleaner.

clxmstaab: I added initial setup for github actions with https://github.com/cweagans/composer-patches/pull/331

sun:
What are the necessary steps to test this on an existing setup using Composer 1?

The community can test manually as long as there is no automated coverage, if everyone understands what to do.

hussainweb:
@sun, [here](https://github.com/hussainweb/drupal-php8-composer2/commit/53ab3b214048a30f40de6b617da4feed5764f395) is the change to composer.json I have on another repository where I am testing both composer 1 and 2. Follow the same steps as that diff on your own project. This will get the patch on your project.

If you want a sample project to test, feel free to use that repo (https://github.com/hussainweb/drupal-php8-composer2). That repo is a little heavy with a lot of dependencies but it works.

Once you have the project setup, test each version of composer simply with `composer install`. After testing with composer 1, run `composer self-update --2` to try with composer 2. To rollback, run `composer self-update --rollback`.

I hope this helps.

danepowell:
I've manually tested this PR with Composer 1 and confirmed that it works as before. I haven't been able to test with Composer 2 since Composer seems to be reading the plugin version from Packagist or something odd. But that's a problem with Composer, not with this PR.

> The "cweagans/composer-patches" plugin was skipped because it requires a Plugin API version ("^1.0") that does not match your Composer installation ("2.0.0"). You may need to run composer update with the "--no-plugins" option.

hussainweb:
@danepowell, I am able to require the plugin with the version constraint `1.x-dev` (that's where the `require` statement is updated). If you are also adding my fork as the alternative repository (see [this commit](https://github.com/hussainweb/drupal-php8-composer2/commit/53ab3b214048a30f40de6b617da4feed5764f395)), that should work as well.

danepowell:
Thanks, I was able to get it working.

I've tested on both Composer 1 and 2 and Mac and Ubuntu, and different versions of Git/Patch, and verified that this pull request works.

grasmash:
Really looking forward to trying out Composer 2 with this project!

cweagans:
Thanks for the PR and for all the help testing!

PR #331, opened 2020-09-24
setup github actions for integration testing

clxmstaab:
on the way to see why composer v2 does not work, I thought it would be usefull to have phpstan coverage and unit testing run with a Github Action.

after this PR is merged, I would propose to add a end-2-end test which should show us the errors the current version of the plugin has with composer v2

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,0 +1,12 @@
+parameters:
+	ignoreErrors:
+		-
+			message: "#^Parameter \\#1 \\$eventName of method Composer\\\\EventDispatcher\\\\EventDispatcher\\:\\:dispatch\\(\\) expects string, null given\\.$#"
+			count: 2
+			path: src/Patches.php
+
+		-
+			message: "#^Call to an undefined method Composer\\\\Package\\\\PackageInterface\\:\\:setExtra\\(\\)\\.$#"
+			count: 1
+			path: src/Patches.php
+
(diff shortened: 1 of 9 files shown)


phpstan-baseline.neon
```diff
@@ -0,0 +1,12 @@
...
+		-
+			message: "#^Parameter \\#1 \\$eventName of method Composer\\\\EventDispatcher\\\\EventDispatcher\\:\\:dispatch\\(\\) expects string, null given\\.$#"
+			count: 2
+			path: src/Patches.php
+
+		-
+			message: "#^Call to an undefined method Composer\\\\Package\\\\PackageInterface\\:\\:setExtra\\(\\)\\.$#"
+			count: 1
+			path: src/Patches.php
+
```
clxmstaab, line 12: these are minor problems, not relevant for composer v2 support.

tests/PatchEventTest.php
```diff
@@ -11,7 +11,7 @@
 use cweagans\Composer\PatchEvents;
 use Composer\Package\PackageInterface;
 
-class PatchEventTest extends \PHPUnit_Framework_TestCase {
```
clxmstaab, line 14: this change was required to get the test running on a more modern phpunit version

clxmstaab:
@cweagans any feedback on this?

clxmstaab:
ping @cweagans

cweagans:
GHA is set up on master. #447 is moving 2.x close to the finish line very quickly. Probably not going to do much with 1.x.

PR #334, opened 2020-11-02
Fix for outdated patch version on FreeBSD

pesc:
Hi

There are many issues about this topic and many tried to get a fix into the repo. 

#184 This is one of the biggest pull request - it got rejected because of many reasons. Many changed to another project to fix it.

So my approach is to use the `--posix --batch` only when the OS is BSD. That shouldn't break anything and makes everyone happy.

[Even Travis is happy](https://travis-ci.org/github/pesc/composer-patches/builds/740790498)

@cweagans Please let me know what you think about this approach. Will then submit a fix for 1.x too


cweagans:
Thank you for this! I missed this when you opened it. This is a very nice, safe way to accomplish what you need. Really appreciate it!

Issue #258, opened 2019-03-15
Backport the removal of PATCHES.txt to 1.x branch


donquixote:
In issue #140 / PR #213, the patch report generation (PATCHES.txt) was removed for the 2.x branch.
A release in this new branch has not happened yet.

I am going to provide a PR that backports this change to the 1.x branch.

donquixote:
The two commits from #213 will be squashed in the backported PR.

alisonjo315:
I applied the change with a PR patch from #259 -- works great, I see no side effects!  (Tested on a Drupal 8 project, with a patch to Drupal core.)

(Side note: I lol'ed when a "PATCHES.txt" file was created in `vendor/cweagans/composer-patches` 😁 -- not a bug at all, obviously, just a funny moment!)

cweagans:
I'm okay with this, but can it be behind a config flag? I know of at least one user that relies on this and I don't want to break their workflow. Definitely okay with this being an opt-out thing though.

awd-studio:
Hey! Would you mind? I made a PR for this.

cweagans:
There is now a config flag for this in 1.x. `main` doesn't have this functionality at all.

PR #337, opened 2020-11-15
#258 Added a flag to skip reporting on applied patches

awd-studio:
According to #258, provided a way to avoid reporting, by using a flag `composer-patches-skip-reporting` with value `true`, or an env-variable `COMPOSER_PATCHES_SKIP_REPORTING`. 

Used similar approach with the `composer-exit-on-patch-failure` flag.

Added a how-to description to the README.

No backward-compatibility brakes.


cweagans:
Thanks for this!

PR #340, opened 2020-11-19
Add support for local patch files

### LINKED ISSUE #339: Locally stored patches in packages
We have a private package with a composer.json like this:

```
{
  "require": {
    "phpoffice/phpspreadsheet": "^1.11",
  },
  "extra": {
    "patches": {
      "phpoffice/phpspreadsheet": {
        "Output only from column D onwards": "patches/column-d.patch"
      }
    }
  }
}
```

The package contains a local file `patches/column-d.patch` and when `composer update` runs, it finds the patch file and applies it.

Now, when we require that private package in a project with `"enable-patching": true` in the extra section, that work well with all the patches that have a fully qualified url (e.g. https://www.drupal.org/files/issues/2019-02-15/3033341-2-domain-pattern.patch) but it doesn't find the local patch file.

With composer 1 we got this working by specifying the patch file name in the package as a relative path from the future root project, but that was (a) not safe and (b) does no longer work with composer 2.

Any idea how we could do that properly?

jurgenhaas:
Turns local patch file uris into absolute paths when the patch file starts with a period.

Resolves #339

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -84,6 +84,21 @@ class Patches implements PluginInterface, EventSubscriberInterface {
   }
 
   /**
+   * Turn local patch file uris into absolute paths.
+   * @param string $install_path
+   * @param array $patches
+   */
+  protected function checkUrls($install_path, &$patches) {
+    foreach ($patches as $package_name => $package_patches) {
+      foreach ($package_patches as $description => $url) {
+        if (strpos($url, '.') === 0) {
+          $patches[$package_name][$description] = implode(DIRECTORY_SEPARATOR, [$install_path, $url]);
+        }
+      }
+    }
+  }
+
+  /**
    * Before running composer install,
    * @param Event $event
    */
@@ -111,6 +126,9 @@ class Patches implements PluginInterface, EventSubscriberInterface {
                 $extra['patches'][$package_name] = array_diff($extra['patches'][$package_name], $patches);
               }
             }
+          }
+          if (!empty($extra['patches'])) {
+            $this->checkUrls($installationManager->getInstaller($package->getType())->getInstallPath($package), $extra['patches']);
           }
           $this->installedPatches[$package->getName()] = $extra['patches'];
         }


src/Patches.php
```diff
@@ -83,6 +83,21 @@ public static function getSubscribedEvents() {
...
 
+  /**
+   * Turn local patch file uris into absolute paths.
+   * @param string $install_path
+   * @param array $patches
+   */
+  protected function checkUrls($install_path, &$patches) {
+    foreach ($patches as $package_name => $package_patches) {
+      foreach ($package_patches as $description => $url) {
+        if (strpos($url, '.') === 0) {
```
ctrl-f5, line 94: I've got it like this, as to not require the patch to start with `./`

```suggestion
        $absolutePath = $install_path . DIRECTORY_SEPARATOR . $url;
        if (file_exists($absolutePath)) {
          $patches[$package_name][$description] = $absolutePath;
        }
```
Either way, both work fine for me!

ctrl-f5 (approved):

cweagans:
This doesn't seem safe from a BC standpoint, as all URLs starting with `.` will a) be transformed to an absolute path (which is going to make composer.lock merges a real headache) and b) all local patches will need to be stored relative to the root of the package that they affect, which is IMO a very severe downside.

I don't think I can merge this as it will break behaviors for _every_ user that's currently using local patches. More importantly, I don't understand why this is needed _at all_ because RemoteFilesystem eventually boils down to just `file_get_contents()`, which supports relative URLs just fine -- can you help me understand how these changes solve the problem presented in #339 where local patches with relative URLs don't work anymore?

jurgenhaas:
The use case is this: we have a composer plugin A which depends on a third party plugin B. Plugin B needs to be patched and for some reason the patch can't be made public. Therefore, plugin A stores the patch locally as `patches/b.patch` and when developing on that plugin, all works just fine and as expected. Now, plugin A becomes a dependency of project C. When running composer update or install in the root of C, the patch `patches/b.patch` won't be found and that's what this PR is trying to solve.

> a) be transformed to an absolute path (which is going to make composer.lock merges a real headache)

Is composer.lock storing the altered patch name? I wasn't aware of that. This could then be resolved by converting the local patch file name from plugin A into a relative path from the project C's composer.lock. That would still work and resolve this issue.

> b) all local patches will need to be stored relative to the root of the package that they affect, which is IMO a very severe downside.

Relative to the root of plugin A in our case - if that's what you mean by "the package that they affect". That's on purpose and I don't see why that's a downside. Can you explore on that?

---

And yes, all this is only required if a patch file can't be published. It may be difficult to justify that to everbody.

jurgenhaas:
I'm closing this PR as we have changed strategy and store those patches in a special repository and load them from there.

PR #343, opened 2020-11-25
Fix #12: Respect composer.lock during package install, but without pa…

### LINKED ISSUE #12: Composer patches only respects composer.json
When running composer install, I expect the composer.lock file to be respected, but instead, the patches defined in composer.json are always used. This isn't critical by any stretch, but it's not consistent with the way composer works.

Proposal:
- Respect composer.json during composer update
- Respect composer.lock during composer install

ghost:
…ckage re-install or re-patch.

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -47,6 +47,10 @@ class Patches implements PluginInterface, EventSubscriberInterface {
    * @var array $patches
    */
   protected $patches;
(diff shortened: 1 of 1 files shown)


cweagans:
Reproducible builds is a priority in 2.x. See #447 for status.

PR #344, opened 2020-11-25
Adds patch options support, with binary option already implemented

alepane21:
Hi, I need the ability to patch some module made in code with \r\l line endings. The only clean way to do it, without using git apply, is to use the binary option on patch. With git apply you can prepare a binary patch with git format-patch, but I cannot do it because our module is not versioned with git. What a wild world we live in!
So my only option was to fork your nice code and try to add the ability to specify some patches as binary.
Let me know if there are things that you don't like and should be fixed, thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,40 @@ Composer [blocks](https://getcomposer.org/doc/06-config.md#secure-http) you from
 
 However, it's always advised to setup HTTPS to prevent MITM code injection.
 
+## Apply binary patches
+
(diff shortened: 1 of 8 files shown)


andypost:
+1 to add support for binary patches

yaroslav-qlicks:
it is cool feature, why it is still not merged?

brucemead:
+1

awm086:
+1

cweagans:
I've made a note to include this in #447.

PR #346, opened 2020-12-03
Report missing packages where patches exist (due to typo in patch package name for example)

driskell:
I've noticed a couple of times where a typo in the patches file meant patches weren't applied at all. Usually you pick it up straight away but I can see it could slip through and you'd not know until you test the impact of the patch.

This adds a last check after install to check for any patches that weren't applied because their package was not installed, and reports it to the console. If exit-on-failure is enabled it will also fail the install (as this is no different to an invalid patch).

It only works at the root level so won't report missing packages that nested dependencies are trying to patch (since the owner of the root package is unlikely to be able to fix it if it did.)

Happy to add the same to the new 2.x branch - I avoided initially as I can see `checkPatches` is disabled so it's likely not ready yet so would be hard to test - maybe when it's ready I can do the same PR.

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -69,6 +69,7 @@ class Patches implements PluginInterface, EventSubscriberInterface {
   public static function getSubscribedEvents() {
     return array(
       ScriptEvents::PRE_INSTALL_CMD => array('checkPatches'),
+      ScriptEvents::POST_INSTALL_CMD => array('checkMissing'),
       ScriptEvents::PRE_UPDATE_CMD => array('checkPatches'),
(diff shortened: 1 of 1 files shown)


driskell:
Hi @cweagans I will have some time to work on this further if there's some bits that need improving. Would be great to hear back. Thanks!

fredden (approved):

cweagans:
I've made a note to include this in #447.

driskell:
@cweagans Thanks! Would be great to cherry pick and adjust 😄

PR #364, opened 2021-06-08
Add support for Promises being returned by uninstall in Composer 2.1

### LINKED ISSUE #363: Packages are getting deleted and not reinstalled since Composer 2.1.0
Since Composer 2.1.0 when running `composer install` we're sometimes seeing packages that need patches getting deleted and not reinstalled.

```
$ composer install
Gathering patches from patch file.
Removing package drupal/redis so that it can be re-installed and re-patched.
  - Removing drupal/redis (1.5.0)
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Nothing to install, update or remove
Deleting app/modules/contrib/redis - deleted
```

Running `composer install` a second time resolves the issue.

```
$ composer install
Gathering patches from patch file.
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Package operations: 1 install, 0 updates, 0 removals
Gathering patches from patch file.
  - Installing drupal/redis (1.5.0): Extracting archive
  - Applying patches for drupal/redis
...
```

Seldaek:
Fixes #363 I hope :)

See also #366 for a version of this PR targetting 1.x

### DIFF
--- a/src/Plugin/Patches.php
+++ b/src/Plugin/Patches.php
@@ -253,6 +253,7 @@ class Patches implements PluginInterface, EventSubscriberInterface, Capable
             }
 
             // Remove packages for which the patch set has changed.
+            $promises = [];
             foreach ($packages as $package) {
                 if (!($package instanceof AliasPackage)) {
                     $package_name = $package->getName();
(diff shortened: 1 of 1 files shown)


mstrelan:
This appears to resolve the issue following the steps in https://github.com/cweagans/composer-patches/issues/363#issuecomment-856316984, although I only applied the changes manually after running `composer require cweagans/composer-patches`. I have not tested this against older versions of Composer.

Seldaek:
Thanks 👍🏻  Older versions return null instead of a promise, so array_filter should just return an empty array and the behavior will not be altered.

See https://github.com/composer/composer/blob/1.10/src/Composer/Installer/InstallationManager.php#L207-L212

cweagans:
Thank you for this, @Seldaek! Really, really appreciate it!!

PR #368, opened 2021-06-11
Fix #12: Respect composer.lock during package NEW install

### LINKED ISSUE #12: Composer patches only respects composer.json

Oscaner:


mxr576:
@Oscaner can you describe what is the problem this patch resolves exactly? I see some comments in your code but tbh, I do not understand it. Maybe @cweagans do.
Providing a detailed description could ensure that others can review this patch and promote it to be merged.

mxr576:
Closes #12

cweagans:
Probably not going to merge this into 1.x, but reproducible builds using a lock file is something that I want to explicitly support in 2.x. Keep an eye on #447 for status.

Issue #244, opened 2019-01-05
Add --ignore-whitespace option when patching


rodrigoaguilera:
I found a case where the plugin fails to patch a file generated with git diff

The git project is https://www.drupal.org/project/nexus
And I attach the patch 

[test-whitespace.txt](https://github.com/cweagans/composer-patches/files/2729357/test-whitespace.txt)

After some digging I discovered that is a whitespace problem and that adding --ignore-whitespace to the patch command makes the patch to be applied.

My temporary solution was to generate the patch with diff instead of git

Can we add that option by default?

Or at least configurable like the patchlevel

Thank you for such a great project :)

kinglozzer:
+1 for this

attisan:
would be a decent default value 👍

seanbeaton:
I've created #370 which implements this as a config option you can add to your composer.json.

cweagans:
See #370 -- IMO, just fix the patch if it's causing issues.

PR #370, opened 2021-06-16
#244 add ignore-whitespace config for patches

seanbeaton:
This patch adds a config option that adds `--whitespace-ignore` to its patch application commands. 

Without that flag, a few of my patches would fail to apply, specifically to some minified js from a drupal contrib module. 

I also updated the readme with instructions on how to use the setting. 

I'd love to add this to the 2.x version as well, though I don't believe there's currently the option to set config on a per-package basis there as there is on the 1.x branch (like the patch-level/patchLevel option). 

It's implemented by choosing between two versions of the command based on the config setting instead of just another argument like the patchLevel one as if it's disabled, it would pass in `''` as an the argument, which gets misinterpreted as the patch filename. If there's a better way to do this, feel free to let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,20 @@ Its useful for packages like drupal/core which packages only a subdir of the ori
 }
 ```
 
+## Ignoring whitespace for patches (--ignore-whitespace)
+
+In some cases, patches won't apply correctly if the original file has some weird whitespace stuff going on. This option can help mitigate those issues.
+
+```json
+{
+  "extra": {
+    "ignoreWhitespace": {
+      "drupal/example-module": true
+    }
+  }
+}
+```
+
 ## Using patches from HTTP URLs
 
 Composer [blocks](https://getcomposer.org/doc/06-config.md#secure-http) you from downloading anything from HTTP URLs, you can disable this for your project by adding a `secure-http` setting in the config section of your `composer.json`. Note that the `config` section should be under the root of your `composer.json`.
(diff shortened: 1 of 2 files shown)


cweagans:
I don't think we need a config option for this. Fix your patch if it's causing issues.

Nico-Schaefer-2111:
@cweagans I don't think that's always possible. In my case I am maintaining the patches within the project repository (git) and deploying them via github actions. 
Therefore git seems to automatically change the line endings of the patch. After that, the patch does not apply anymore on the target system, and the only way around that else than fix the vendor file line endings or the patch line endings after deployment, which is really annoying to maintain for every specific customer repository if they may or may not use some of these packages or not.

rommelfreddy:
@cweagans can confirm @Nico-Schaefer-2111 's issue.

this option would be really helpfull

floviolleau:
Hi,

I face exactly to the same issue as @rommelfreddy .

Is it possible to reopen this PR?

Thanks

Nico-Schaefer-2111:
@cweagans this is obviously a necessary feature, please reopen the PR

Nico-Schaefer-2111:
@cweagans  in our case it was the same as with @rommelfreddy 

Sometimes we cannot control what externals put in their code or how they configured the line endings and that's why it would be very useful to support that case.

rommelfreddy:
you are absolutely right. 

But your answer still does not respect the problems with external developers, which are not able to work with correct line-endings.

The author of the PR already added a flag to make sure, not all patches are runs with `--ignore-whitespace`. this flag has to be enabled for each package.

So this is *absolutely safe*, and will *not break* any behaviour for existing setups.

So, I kindly ask again to review the PR again, and think about the votes/comments in this PR again.

cweagans:
This PR is against 1.x. Nothing else is going in to 1.x. Upgrade to 2.x.

I gave you some options for 2.x as well. A dedicated switch for this specific flag isn't going in, but I'd accept a PR that adds the ability to specify extra CLI flags on a per-patch basis.

PR #375, opened 2021-06-29
Use the -i flag for input file, and do not pipe via stdin.

cafuego:
Turns out that under docker (on GitHub Actions for one) stdin does not
behave the way we would like. Although patching suceeds, the command
returns an error code and composer exits. Using the `-i` option to
specify the input file resolves this issue and is not less compatible.


cweagans:
Do you happen to know if this is equally compatible across all of the different variations of `patch`? BSD vs GNU vs Apple vs Busybox?

cafuego:
It definitely works with the GNU and Apple versions of patch. The googled source code of a 6 year old busybox reckons that it also supports -i, as does a search of the freebsd man (1) patch.

cweagans:
I've made this change in #447 and credited you as the author of the commit.

PR #163, opened 2017-10-21
Validate Patch Checksum

gapple:
Still a work in progress, but intended to supersede #24

Issue #354, opened 2021-02-12
Security: Deprecate "enable-patching" option


metalbote:
Correct me, if I'm wrong but if i enable "enable-patching" in my project, patches from all my required packages are applied. 

Is there a possibility to restrict this behavior to a certain set of package, for examples i just want to get those patches apllied from myvendor/x, and not from all packages my project installs?

This would have several benefits, 
a) i have a project which could be used just as a quick starter, and all of patches, maintaining etc could be done in a required package.
b) I do not have to worry about security by patches being applied from vendor/xyz, only those i explictly added in the allowed patching module.

acha5066:
I also think this would be a good feature. I want to provide patches to certain modules in a utility module we use internally on a couple of projects. However, I don't want to allow every dependency to provide patches.

mxr576:
I wanted to open a new issue but I am glad that I am not the only one who finds "enable-patching" option is a slippery sloop and a possible security issue, namely an open door for supply chain attacks.

I put together [a small POC](https://github.com/mxr576/composer_patches_sca_via_external_deps) that demonstrates how one small dependency in a dependency tree by _ANY_ vendor can sneak in code to your application when you have `enable-patching: true;` in your root composer.json.

The story that my dummy POC would like to tell that
* [module1](https://github.com/mxr576/composer_patches_sca_via_external_deps/tree/main/web/modules/custom/module1) was developed by our team
* [module2] (https://github.com/mxr576/composer_patches_sca_via_external_deps/tree/main/web/modules/custom/module1) is an external dependency of our code
* which also depends on a [3rd party library](https://github.com/mxr576/composer_patches_sca_via_external_deps/tree/main/web/modules/custom/malicious_dependency) that gone rouge and either ships a previously trusted patch with malicious content or just added a new patch with malicious code, it does not matter

(Please ignore that these are local packages, they could be any `anyvendor/awesomepackage` in `vendor`.)

I understand why developers would like to deliver/bundle patches with libraries and if we need a similar feature my proposed alternative would be something like Composer 2.2.0 introduced with [allow-plugins](https://getcomposer.org/doc/06-config.md#allow-plugins). Let's deprecate "enable-patching" and replace with for example an `apply-patches-from` list that contains all those _trusted_ packages that can apply patches on a project. Example:

```json
{
    "name": "mxr576/my-project",
    "type": "project",
    "minimum-stability": "stable",
    "extra": {
        "apply-patches-from": {
[...]

Issue #347, opened 2020-12-16
Patches from drupal.org merge request URLs are dangerous?


JohnAlbin:
`composer-patches` allows developers to specify patches to download and use via HTTP/S, which is a super useful feature that we use all the time!

This is totally safe to do when the resource at the end of that URL is static (unchanging). For example, a URL to an actual `.patch` file on a website is safe to use as long as the website owner is not malicious.

The new drupal.org merge request URLs are not safe, however. For example: https://git.drupalcode.org/project/drupal/-/merge_requests/121.diff

Anyone with commit access to Drupal issues can change the code on that the URL above references. That’s a malicious code injection vector. 

The [docs on drupal.org describing how to create a patch file](https://www.drupal.org/drupalorg/docs/gitlab-integration/issue-forks-merge-requests#s-patch-files-for-use-with-composer) from a merge request specifically mention that the URL is static, but the code is not:

> WARNING: The URL for a Merge Request patch file will be named after the MR and will not change as more and more commits are being added in the MR. The contents of the patch file will change however.

However, those docs assume that a developer would review the patch before applying it manually. They don't take into this feature of `composer-patches` that allows you to automatically apply whatever patch is at the end point.

> Anyone with commit access to Drupal issues can change the code on that the URL above references.

It wouldn't be that hard for a malicious user to pretend to be an open-source developer. If they knew a particular website was using a merge request URL for production builds, they could target that site. Or, theoretically, they could insert some phone-home code that exploits and then contacts the malicious user afterwards.

Obviously, I haven't tried this yet. Maybe I'm missing something and it's not even theoretically possible. But I'd love to see some eyeballs on this to make sure a scenario like this isn't possible.

cweagans:
I think you're right -- this probably warrants at least a warning for people using that kind of URL. It may also be a good idea to see if Gitlab has some way to pin a merge request URL to a specific commit or something (so that you can say I've reviewed the MR patch at this specific point in time and that's what I'm going to apply to my codebase). That seems like a really specific feature that Gitlab may not have thought to implement though.

Longer term, this plugin will start storing SHA-256 hashes of all patches in the lock file (if not manually specified, then the file will be hashed when the patch is first applied), so if something changes on the server side that changes that patch contents, there will be an error.

jonathanjfshaw:
[...]

PR #388, opened 2022-01-02
Add support for patch checksums

ceesgeene:
As the status of version 2 is not clear to me, this is a PR against 1.x instead of the master branch. It would be nice if we could have this feature in version 1.

This PR is based upon #163 . That PR no longer matches. Biggest difference is that this PR keeps using the key as description. Also instead of hash, the term checksum is used.


mxr576:
First of all, I am thrilled that there is some movement on this topic :tada: How Composer Patches could be leveraged in supply chain attacks was one of those problems that concerned me the most in the past year(s), but my focus was always shifted so I could not work on these problems. Also, @cweagans made it quite clear on [Twitter](https://twitter.com/cweagans/status/1430664598319890432) that no matter how much the PHP community (not just Drupal developers) [depends on this package](https://packagist.org/packages/cweagans/composer-patches/stats), it is not minimally maintained, and _there is no problem with that_. Although, that made me realize that a community effort and support are needed to provide at least the minimum necessary care to this valuable library. Something similar that was formulated around https://github.com/php-tuf/php-tuf.

Maybe that will happen now, and I would be glad if I could be part of that effort.

mxr576:
As I said in https://github.com/cweagans/composer-patches/issues/347#issuecomment-1004603765 I am really happy that there is some movement around this topic, so thanks for this PR.

What I like about this approach that it is _simple and straightforward_. What I do not like that it is opt-in feature and security must not be optional. So I started to think about how we can have this security feature enabled by default.

We have two significantly different scenarios, when a patch is installed from a local source or from a remote URL. We only have problems with remote patch sources because in general, we cannot trust that the content of a patch does not change on a remote. If a patch is served from the local filesystem that was potentially downloaded _and verified_ by a human, hopefully also passed on a code review and it is part of a VCS repo. Why do I think it is essential to differentiate these two scenarios? The reason is that if hash based protection is enabled by default and it would also be applied on local patches, then I think it could become a bottleneck because the content of those patch files can rightfully change anytime and no additional steps (like running `composer update-composer-patch-hash ./path-to.patch`)  should be required for updating a local patch file.

So if we would like to auto-generate hashes (with a preferred algorithm) of newly added _remote_ patches, how can we do that?

When I saw these manually generated hashes, it made me remember again that with Composer 1, all applied patches [were listed in composer.lock also](https://github.com/DrupalOpenLearning/drupalopenlearning/blob/7a46c10e36fce58afc2c2320663f523e8457fa54/vendor/composer/installed.json#L1926). I am not sure how and why that changed with Composer 2 but my opinion is that auto-generated hashes should be in composer.lock, just like [hashes of downloaded packages](https://github.com/DrupalOpenLearning/drupalopenlearning/blob/7a46c10e36fce58afc2c2320663f523e8457fa54/vendor/composer/installed.json#L1908).

When a new patch is added and `composer install` or `composer update` is executed, then Composer Patches plugin could generate hashes for remote patches that did have hashes before. This also implies that the URL of a remote patch is considered unique, which was avoided here as a design decision, but is it a good idea? When a diff of a Gitlab MR is referenced and the MR changes what we want is make the build failed because it is a _mutable_ (non-static) reference.

In which circumstances we need hashes in `composer.json`? I can imagine one common scenario when patches are distributed via libraries/modules ([example 1](https://github.com/nwuatmath/drupal/blob/ce1d6481f4f8e075977cc053714a252f36a64c37/lightning-8.x-3.200/modules/contrib/lightning_layout/composer.json#L79-L90)) by leveraging the `enable-patching: true` feature of Composer Patches (which is another [open door for supply chain attackes](https://github.com/cweagans/composer-patches/issues/354#issuecomment-1004759055)). In that scenario, the package provider probably also would like define a hash for every  _remote_ patch "suggested" by the library so whenever patches are installed on a project, Composer Patches could verify that the content of a proposed patch is the same as it was when it had been added.

What are others thinking about these ideas?

(Possibly the easiest and safest option would be only supporting local patches :grin: )

cweagans:
Thank you for opening this -- I'm swamped but will review ASAP. Thank you @gapple for the original PR as well.

One (hopefully) quick thing: let's just pick a hashing algo and only support that one. sha256 seems like a good candidate, but I'm open to other options. I'd prefer to reduce the number of possible configurations + limit the scope of this change as much as possible.

ceesgeene:
I agree on limiting the scope as much as possible; refactored the PR so it only supports sha256 checksums.

bbrala:
This really would be awesome to get moving. This would be an awesome step forward, even though there is probable next steps to make this even more automated.

ceesgeene:
@cweagans, haven't found the time yet, or is something regarding the PR holding you back?

This is an attempt to push this pull request  :)

korkhau:
The time for merge is come.
@cweagans please do :)

cweagans:
#447 has this.

PR #391, opened 2022-01-06
PHP 8.0: Allowing tests to run under php 8.0

ChrisB9:
Many packages in the composer.json (require-dev) are not php 8.0 compatible.
I upgraded all packages to check whether this library can run reliably under the newer php version.
This though does make it very hard to merge it with the master branch as significant changes had to be applied to grumphp and codeception. Other packages are not 5.3 compatible.

I did not come up with a solution yet to allow both 5.6+ and 8.0 compatibility, therefore this PR currently just shows this package works under php 8.0.


danepowell:
I think we simply need to drop support for PHP 7 and below in the 2.x release, since these PHP versions are long EOL anyway.

The 1.x branch may continue to support them but only on a best-effort basis. It's too hard to keep old PHP and Composer versions running in CI.

PR #392, opened 2022-01-24
Add a method to tell if a package has been patched

phenaproxima:
For the purposes of Drupal's auto-updates initiative, it would be really useful to us to have an API method in this plugin which can tell us if an installed package has been patched. We could just check the presence of the `extra.patches_applied` key -- and indeed, we do -- but we'd rather the plugin exposed an API for this, so we don't have to rely on internal keys. So this PR adds a `Patch::isPackagePatched()` method for this purpose.

See related discussion in https://www.drupal.org/project/automatic_updates/issues/3252299.

### DIFF
--- a/src/Patch.php
+++ b/src/Patch.php
@@ -2,6 +2,8 @@
 
 namespace cweagans\Composer;
 
+use Composer\Package\PackageInterface;
+
 class Patch implements \JsonSerializable
 {
     /**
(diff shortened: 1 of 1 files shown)


cweagans (approved):

cweagans:
Note that if you want this in a release any time soon, this will also need to go into `1.x`

PR #396, opened 2022-02-19
Backward compatibility of #334 for 1.x

pesc:
Hi

Some time ago, I've created the pull request: #334 which was merged to master branch. The problem is, the most used version is 1.x. So we need to fix it for this version too. Cherry-picking wasn't a option, because the #334 version only works for PHP 7.2+.

So my approach is based on the new `PHP_OS_FAMILY` (PHP 7.2+) mixed with older `PHP_OS` detection and the help of this issue: https://github.com/sebastianbergmann/environment/issues/21

@cweagans What do you think?

manual of `patch`:
https://man.netbsd.org/patch.1
https://man.openbsd.org/patch.1
https://www.freebsd.org/cgi/man.cgi?patch(1)

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -409,7 +409,7 @@ class Patches implements PluginInterface, EventSubscriberInterface {
       // --no-backup-if-mismatch here is a hack that fixes some
       // differences between how patch works on windows and unix.
       $patch_options = '--no-backup-if-mismatch';
-      if (PHP_OS_FAMILY == 'BSD') {
+      if (PHP_OS_FAMILY == 'BSD' || PHP_OS == ('FreeBSD' || 'NetBSD' || 'OpenBSD')) {
           $patch_options = '--posix --batch';
       }
       foreach ($patch_levels as $patch_level) {
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -395,10 +395,16 @@ class Patches implements PluginInterface, EventSubscriberInterface {
     // In some rare cases, git will fail to apply a patch, fallback to using
     // the 'patch' command.
     if (!$patched) {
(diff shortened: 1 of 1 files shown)


pesc:
Closes #182
Closes #184

pesc:
@cweagans any thoughts on this PR?

mfb:
I tested this on FreeBSD and it didn't actually work for me - it seems to run `patch '-p4' '--posix --batch' -d 'web/core' < '/tmp//6252865cb0bbf.patch'`

which results in patch: unrecognized option `--posix --batch'

I guess because escapeshellarg(), which adds single quotes, is run on the options as if they were one argument.

mfb:
The approach in #402 would work here too - building an array of options which can be escaped individually.

pesc:
@mfb Yeah could be, but actually you're talking about a different issue. Without #334 or this #396 it wouldn't even work on FreeBSD. 
I'll have a look on the changes in #402 and see if I can update this PR to the 1.X branch

mfb:
Well, the point is that #334 and this PR #396 don't yet work on FreeBSD, but I think we're getting close - #402 is now able to get the master branch working on FreeBSD, and we need similar code for this PR on the 1.x branch.

pesc:
@mfb Ok, I see the issue. I'll wait until #402 is merged and will then adjust the backport

cweagans:
#447 has explicit support for BSD patch. Please take a look at `src/Patcher/*` - let me know if that works for you.

pesc:
hmm @cweagans...in this PR the goal was to backport the BSD Patch feature to 1.X but #447 is only for 2.X

Issue #401, opened 2022-03-28
Paramters '--posix --batch' in FreeBSD failed


RicardoSette:
https://github.com/cweagans/composer-patches/blob/55c428957e791404ecfbe63c2bd9b43f57721550/src/Plugin/Patches.php#L446

Hi, in FreeBSD this line causes an error because both parameters are executed inside single quotes, see verbose after "composer update -v":

```
patch '-p1' '--posix --batch' -d '/usr/local/www/vendor/yiisoft/yii2' < '/usr/local/www/patches/yiisoft-yii2-widgets-pjaxasset-php.patch'
patch: unrecognized option `--posix --batch'

usage: patch [-bCcEeflNnRstuv] [-B backup-prefix] [-D symbol] [-d directory]
             [-F max-fuzz] [-i patchfile] [-o out-file] [-p strip-count]
             [-r rej-name] [-V t | nil | never | none] [-x number]
             [-z backup-ext] [--posix] [origfile [patchfile]]
       patch <patchfile

```

Where only the 'batch' parameter would suffice, see a functional example of the command

```
% patch '-p1' '--batch' -d '/usr/local/www/vendor/yiisoft/yii2' < '/usr/local/www/patches/yiisoft-yii2-widgets-pjaxasset-php.patch'
Hmm...  Looks like a unified diff to me...
The text leading up to this was:
--------------------------
|--- /dev/null
|+++ /widgets/PjaxAsset.php
--------------------------
Patching file widgets/PjaxAsset.php using Plan A...
Patch creates file that already exists, Assuming ReversedHunk #1 succeeded at 19.
done
```

Both parameters also work, but each must be inside its own single quotes, see example:
```
% patch '-p1' '--posix' '--batch' -d '/usr/local/www/vendor/yiisoft/yii2' < '/usr/local/www/patches/yiisoft-yii2-widgets-pjaxasset-php.patch'
Hmm...  Looks like a unified diff to me...
The text leading up to this was:
--------------------------
|--- /dev/null
|+++ /widgets/PjaxAsset.php
--------------------------
Patching file widgets/PjaxAsset.php using Plan A...
Patch creates file that already exists, Assuming ReversedUnreversed (or previously applied) patch detected!  Ignoring -R.Hunk #1 succeeded at 19.
done
```

The important thing is to have only one parameter inside each single quote.

> **Wrong**: '--posix --batch'
> **Correct**: '--posix' '--batch'

cweagans:
[...]

PR #402, opened 2022-04-03
Fix Issue  #401 - correctly apply multiple command line parameter flags for FreeBSD

arcanumbridge:
This is a fix for issue #401 to individually escape the flags to patch.


src/Plugin/Patches.php
```diff
@@ -503,7 +503,13 @@ protected function executeCommand($cmd)
         $args = func_get_args();
         foreach ($args as $index => $arg) {
             if ($index !== 0) {
-                $args[$index] = escapeshellarg($arg);
+                if(is_array($arg)) {
```
mfb, line 506: I tested this PR on FreeBSD and got an array to string conversion error, due to attempting to append a string to an array. Setting the arg to empty string so it can be appended to should resolve the error:

```suggestion
                if(is_array($arg)) {
                    $args[$index] = '';
```

mfb (changes requested):

mfb:
👍 Tested on FreeBSD, it now runs `patch '-p2'  '--posix' '--batch' -d 'web/core' < '/tmp//6253330a8bb7f.patch'`

(I do see a mix of code styles e.g. `foreach(` and `if(` vs `foreach (` and `if (` - not sure if this package has code style preferences or not :)

grappler (approved):
I have just tested the patch and can confirm it is working as expected.

An alternative to the nested if statement would be to use `continue` instead, like I did here: https://github.com/wearerequired/composer-patches/commit/d1a08bd16b2e7a0ac318d9ae0261092f496a3601

mfb (approved):
Tested again on FreeBSD and still working fine - it executes `patch '-p1'  '--posix' '--batch'` rather than the broken `patch '-p1' '--posix --batch'`

Now I guess we need an actual maintainer to review :)

arcanumbridge:
@cweagans @danepowell can we get a maintainer to review and merge this?  This fixes FreeBSD support.

mfb:
@arcanumbridge I found some other issues mentioning non-GNU patch also being broken on macOS. I'm not sure if macOS uses "BSD" as `PHP_OS_FAMILY` or something else?  If "BSD," then hopefully this PR also fixes macOS. If something else, then maybe this code should be checking whether or not patch is GNU patch rather than looking at the operating system family.

mfb:
Realized I could just check the php source code :) Looks like macOS uses "Darwin" as the `PHP_OS_FAMILY`, so this PR won't help there. But, could we tweak the logic to check if patch --version is not GNU patch?  See #426 for a way to check this.  And if so, apply the BSD options.

mfb:
@arcanumbridge should we roll this back to the version that I tested, with BSD OS-specific logic?

cweagans:
Can you please take a look at #447 (in particular, `src/Patcher/*`) and let me know if I've captured what needs captured?

cweagans:
(for context, 2.x is essentially a rewrite and it breaks a _ton_ of things out of the main plugin so that they're individually testable)

danepowell:
I don't have time to test it this weekend, but I like #447 and it looks like it should fix the issue. I like that the patch providers are more structured and test the patch version directly instead trying to infer it based on OS info.

arcanumbridge:
@cweagans I've made a few comments regarding some consistency of using ->patchTool() and testing for gpatch, but I think it essentially captures what we want.

cweagans:
Great! I'm going to close this then and focus on getting 2.x out the door as quickly as possible.

(also thank you for the review -- super helpful)

PR #415, opened 2022-07-13
Communicate that patch failed due to missing 'patch' command

Chekote:
## Problem
When attempting to use this package on a system that does not have the "patch" command installed, you get a vague error message:

> Could not apply patch! Skipping. The error was: Cannot apply patch ######

## Cause
I spend a long time trying to work out exactly why my patch was not applying. Thinking that it was a problem with the patch file itself. But eventually, I discovered that the package is assuming the system has a "patch" command.

##
Check if the patch command is present and can be run before trying to use it. If that fails, then clearly communicate to the user what the problem is so they can resolve it:

> Could not apply patch! Skipping. The error was: Failed to run the "patch" command. Is it installed on the system?

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -404,6 +404,10 @@ class Patches implements PluginInterface, EventSubscriberInterface {
     // In some rare cases, git will fail to apply a patch, fallback to using
     // the 'patch' command.
     if (!$patched) {
+      if (!$this->executeCommand('patch -v')) {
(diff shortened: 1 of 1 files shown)


Chekote:
howdy @cweagans 👋 

Would you mind taking a look at this PR? 🥺 

Thanks 🙇

cweagans:
#447 takes care of this for 2.x.

PR #417, opened 2022-08-12
Allow to load patch from specific package

johannes85:
This changes allows to load the patch from a specific packageb by defining its location in the following way: `//package/name//patch/to/patch`

For example:
```
"extra": {
        "patches": {
            "xp-framework/core": {
                "Safer Objects::stringOf": "//my/package//src/main/patches/xp-framework-core-src-main-php-util-objects-class-php.patch"
            }
        }
    },
```

I implemented this change because when you define a local patch file in a package like "patches/patch1.patch"  the installation of this patch only works when executing composer in the package itself. When the package is a dependency of a project( with "enable-patching"), the patch file can't be found since the current root directory is another one.

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -368,6 +368,25 @@ class Patches implements PluginInterface, EventSubscriberInterface {
    */
   protected function getAndApplyPatch(RemoteFilesystem $downloader, $install_path, $patch_url, PackageInterface $package) {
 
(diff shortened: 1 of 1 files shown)


cweagans:
Probably not going to do this in 1.x, and 2.x doesn't have the idea of resolving patches from dependencies.

PR #418, opened 2022-08-26
Add sample commands for quick installation method

MurzNN:
I've added a block to documentation with sample commands for quick installation method. 

Very often I have temporary projects when want to install `cweagans/composer-patches` quickly, without manually editing my composer json, so use such commands.

So I think it will be useful for other package users too!


cweagans:
I've made a note to include something like this in #447.

PR #422, opened 2022-10-30
Add installedPatches property to fix PHP 8.2 deprecation 1.x version

seamuslee001:
This is the 1.x version of #409

PR #425, opened 2022-12-08
fix: add non-dynamic property for $installedPatches

nesl247:
This removes the deprecation notice about creation of a dynamic property.


nesl247:
Closed as duplicate of #422

PR #426, opened 2022-12-09
Check if GNU patch is used

### LINKED ISSUE #326: Requires GNU patch 2.7 for correct behavior, but `composer-patches` does not warn the user about this!
> @cweagans holy shit, the need for `brew install gpatch` took me many hours to figure out, because for a particular project TravisCI builds were succeeding but local builds were failing. 😬  Can we make `cweagans/composer-patches` detect the `patch` version, and explicitly warn the user that any version below 2.7 may result in silent failures?

_Originally posted by @wimleers in https://github.com/cweagans/composer-patches/issues/172#issuecomment-674860594_

The solution:

> `brew install gpatch` (you'll need Homebrew installed for that to work (https://brew.sh))

_Originally posted by @cweagans in https://github.com/cweagans/composer-patches/issues/172#issuecomment-587147692_

alfredbez:
fixes #326

### DIFF
--- a/src/Plugin/Patches.php
+++ b/src/Plugin/Patches.php
@@ -443,6 +443,14 @@ class Patches implements PluginInterface, EventSubscriberInterface, Capable
             // differences between how patch works on windows and unix.
             $patch_options = '--no-backup-if-mismatch';
             if (PHP_OS_FAMILY == 'BSD') {
+                $patchVersion = '';
+                $this->executor->execute('patch --version', $patchVersion);
+                if (strpos($patchBinaryVersion, 'GNU patch') === false) {
+                    throw new \Exception(
+                        "Please make sure to have GNU patch installed, see "
+                        . "https://github.com/cweagans/composer-patches/wiki/Install-%22patch%22-binary"
+                    );
+                }
                 $patch_options = '--posix --batch';
             }
             foreach ($patch_levels as $patch_level) {


src/Plugin/Patches.php
```diff
@@ -443,6 +443,14 @@ protected function getAndApplyPatch(RemoteFilesystem $downloader, $install_path,
             // differences between how patch works on windows and unix.
             $patch_options = '--no-backup-if-mismatch';
             if (PHP_OS_FAMILY == 'BSD') {
+                $patchVersion = '';
+                $this->executor->execute('patch --version', $patchVersion);
+                if (strpos($patchBinaryVersion, 'GNU patch') === false) {
+                    throw new \Exception(
```
mfb, line 449: The intention here should be for non-GNU versions of patch to actually work! :)  Please review the PR at #402 where we try to get the non-GNU patch working (with a pretty trivial fix).

mfb (changes requested):

src/Plugin/Patches.php
```diff
@@ -443,6 +443,14 @@ protected function getAndApplyPatch(RemoteFilesystem $downloader, $install_path,
             // differences between how patch works on windows and unix.
             $patch_options = '--no-backup-if-mismatch';
             if (PHP_OS_FAMILY == 'BSD') {
```
danepowell, line 445: This won't work reliably because PHP_OS_FAMILY reflect the _build target_ for the running PHP version, not the current hosting environment. In my case on MacOS Ventura, for whatever reason, PHP_OS_FAMILY is Linux.

mfb: I think what you're saying is the PHP documentation is wrong: When it says "The operating system family PHP was built for" what they actually mean is "The operating system family PHP was built on"?  If so, uhg..

So maybe looking at patch --version is correct after all, although all the logic needed there might not be pretty.

danepowell: Yeah the only semi-reliable way to detect runtime OS is [php_uname](https://www.php.net/manual/en/function.php-uname.php), but even that will fall back to displaying the _build_ OS in some situations. I found the [phposinfo](https://github.com/loophp/phposinfo) library to do a pretty good job of compensating for this by using a combination of php_uname, PHP_OS_INFO, and shell commands to verify the OS version.

It's obviously better to avoid any conditional logic based on the OS in the first place.

cweagans:
#447 takes care of this

PR #428, opened 2022-12-19
Add automated tests, require PHP 8

danepowell:
- Travis CI is no longer reliable or secure (or working at all in the case of this project). Use GitHub Actions for automated tests instead.
- Require PHP 8 for the 2.x release (it's just too hard to continue to support PHP 7 for local dev and CI)

I still need to fix the unit and acceptance tests to support Composer 2.

@cweagans any qualms about requiring PHP 8 for the 2.x release or switching to GitHub Actions (free for public projects like this)?


cweagans:
> any qualms about requiring PHP 8 for the 2.x release or switching to GitHub Actions (free for public projects like this)?

Nope! I'm generally okay with dropping support for EOL PHP versions -- even in 1.x, though that would likely need some planning/comms.

PR #440, opened 2023-01-20
Fix #408: Support proxies

### LINKED ISSUE #408: Support alternative transport or proxy
When `ALL_PROXY`, `HTTPS_PROXY`, `HTTP_PROXY` are present in the environment, composer can work with them, but patching stops working, below is an example error message:

```
  - Installing drupal/commerce_stripe (dev-1.x 66b6b4d): Cloning 66b6b4de23 from cache
  - Applying patches for drupal/commerce_stripe
    https://www.drupal.org/files/issues/2021-10-04/commerce_stripe-allow-off_session-payment-intents-for-recurring-payments-3171408-34.patch (Allow off_session payment intents for recurring payments)
   Could not apply patch! Skipping. The error was: The "https://www.drupal.org/files/issues/2021-10-04/commerce_stripe-allow-off_session-payment-intents-for-recurring-payments-3171408-34.patch" file could not be downloaded: Failed to open stream: Unable to find the socket transport "socks5" - did you forget to enable it when you configured PHP?

                                                                                                                                                                                 
  [Exception]                                                                                                                                                                    
  Cannot apply patch Allow off_session payment intents for recurring payments (https://www.drupal.org/files/issues/2021-10-04/commerce_stripe-allow-off_session-payment-intents  
  -for-recurring-payments-3171408-34.patch)!
  ```
  
Probably, this is caused by http client isn't configured to support proxy (`socks5`), the solution is to find the code, and copy it from composer.

skyred:
See https://github.com/cweagans/composer-patches/issues/408

## Motivation:

Composer supports socks5 proxy, we want the same support in patching.

## Approach:

We are currently using a streamer [RemoteFilesystem.php](https://github.com/composer/composer/blob/main/src/Composer/Util/RemoteFilesystem.php), which doesn't support `socks5` or `socks5h` because of protocol limitation. However, we don't really need a streamer, we simply want to download the patch file and move on. So, using a downloader makes sense.

### DIFF
--- a/src/Plugin/Patches.php
+++ b/src/Plugin/Patches.php
@@ -24,7 +24,7 @@ use Composer\Script\Event;
 use Composer\Script\ScriptEvents;
 use Composer\Installer\PackageEvent;
 use Composer\Util\ProcessExecutor;
-use Composer\Util\RemoteFilesystem;
+use Composer\Util\HttpDownloader;
 use cweagans\Composer\Capability\ResolverProvider;
 use cweagans\Composer\PatchCollection;
 use cweagans\Composer\Resolvers\ResolverBase;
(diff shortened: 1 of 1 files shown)


noot-to-the-noot:
@danepowell can you look over this? I've noticed that @cweagans appears to not look into this repository frequently and hope you have the rights to merge here?
Much appreciated!

danepowell (approved):

noot-to-the-noot:
Thanks a lot!

PR #443, opened 2023-01-26
Fix #408: Support proxies

### LINKED ISSUE #408: Support alternative transport or proxy

danepowell:


totten:
@danepowell @cweagans  This probably broke compatibility between `composer-patches@1.x` and `composer@1.x`.

```
Fatal error: Uncaught Error: Class 'Composer\Util\HttpDownloader' not found in
/tmp/assetplg-2a394c08d4f8ae329b1d7eeda1a118d5/vendor/cweagans/composer-patches/src/Patches.php:312
```

It's not a big deal for me personally.  I run some integration-tests with different plugins+composer builds, and this got dinged with `composer@1.10.26`. Tightening the constraints is sufficient for my tests.

But [the `composer.json` for 1.x does declare compatibility with older composer APIs](https://github.com/cweagans/composer-patches/blob/f403f9401adb429af08acb4ee414adbef9c263b3/composer.json#L18-L21).  Maybe consider an updated declaration or else add a fallback `RemoteFilesystem`?

danepowell:
I thought Composer 1.x had been EOL for some time now. I don't think libraries should continue to support it.

cweagans:
Agreed. If we don't support it, we should probably say so though.

PR #445, opened 2023-02-02
Fix #423: Patching stalls on Ventura

### LINKED ISSUE #423: Can't patch after update to MacOS Ventura
I am not sure if this is actually an issue of this package or if it is caused somewhere else. Does anyone else has the problem, that while applying a patch (on MacOs Ventura) the process runs into a timeout? This happens only since the recent update.

After hanging on "File to patch" the following error is thrown:

```
File to patch: 
   Could not apply patch! Skipping. The error was: The process "patch '-p1' --no-backup-if-mismatch -d '/Users/mrl/Repositories/m2-composer-base/vendor/amasty/module-gift-card-account' < '/Users/mrl/Repositories/m2-composer-base/patches/composer/Amasty_GiftCardAccount_Fixes.patch'" exceeded the timeout of 300 seconds.
Downloading https://packagist.org/downloads/
[201] https://packagist.org/downloads/

                                                                                                                      
  [Exception]                                                                                                         
  Cannot apply patch Fixed amasty broken namespaces (AMO-783) (patches/composer/Amasty_GiftCardAccount_Fixes.patch)!  
```

danepowell:
tl;dr: newer `patch` versions on Mac appear to require the `--batch` option, otherwise they stall awaiting input of a filename to patch.

I tested with Ventura and gpatch, both have the `--batch` option so it should be safe to use universally, but to be extra safe I include a check that it's supported.

Note that this would have _almost_ worked with the existing code, except for two bugs:

- PHP_OS_FAMILY is unreliable and returns the OS of where PHP was built, not where it runs, and in any case should be checking for `Darwin` on mac
- `--posix --batch` was getting quoted as a single flag and causing `patch` to error.

Note that to test this properly, you'll need to either have a patch that fails Git patching (so it falls through to the patch utility), or you can comment out / delete the Git patch methods.

### DIFF
--- a/src/Plugin/Patches.php
+++ b/src/Plugin/Patches.php
@@ -447,12 +447,16 @@ class Patches implements PluginInterface, EventSubscriberInterface, Capable
         // In some rare cases, git will fail to apply a patch, fallback to using
         // the 'patch' command.
         if (!$patched) {
+            // Patch on macOS Ventura expects file input unless you use --batch.
+            // @todo use --batch always, if we can determine that all versions
+            // of patch support it.
+            $batch = $this->executeCommand('patch %s', '--batch');
             // This is a workaround for the outdated patch version on BSD
             // systems which doesn't support this option -> use posix then.
             // --no-backup-if-mismatch here is a hack that fixes some
             // differences between how patch works on windows and unix.
             foreach ($patch_levels as $patch_level) {
-                if (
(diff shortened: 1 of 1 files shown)


cweagans:
Closing per https://github.com/cweagans/composer-patches/issues/423#issuecomment-1427372947 but we can reopen if needed.

PR #470, opened 2023-02-12
Add composer and io to all events + emit event before throwing exception when patch does not apply

cweagans:
## Description

Relates to/closes #issueID.

## Related tasks

- [ ] Tests have been added


cweagans:
cc @waleedq @moerazem @Natshah

PR #471, opened 2023-02-12
Add new patcher: git 'init' patcher

cweagans:
## Description

`git` has historically been the most reliable mechanism for applying patches. Unfortunately, it doesn't work at all when the target directory isn't a git repo. This pull request adds a new patcher that creates a fake (minimal) git repository in the target directory, uses git to apply the patch, and then cleans up the fake git repo.

## Related tasks

- [x] Tests have been added

## Other notes

@deviantintegral I'm very curious about your thoughts on this one since you did most of the original debugging around why `git` was misbehaving. (`composer install` then `./vendor/bin/codecept run tests/acceptance/ApplyGitPatchFromWebUsingInitCept.php --debug` if you want to play with this locally)

If this approach works, I'm really considering https://github.com/cweagans/composer-patches/pull/472 as a path forward. Git handles a _lot_ of things that people have asked for over the year. It's also a _lot_ easier to support -- "Go install `git`" is easier to communicate than "Go install `patch`... no, the `patch` that came with your system is unusable and you need to install this other thing that is named exactly the same but provides different functionality"

### DIFF
--- a/src/Patcher/GitInitPatcher.php
+++ b/src/Patcher/GitInitPatcher.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace cweagans\Composer\Patcher;
+
+use Composer\IO\IOInterface;
+use cweagans\Composer\Patch;
+
+class GitInitPatcher extends GitPatcher
+{
+    public function apply(Patch $patch, string $path): bool
+    {
+        // If the target is already a git repo, the standard Git patcher can handle it.
+        if (is_dir($path . '/.git')) {
+            return false;
+        }
+
+        $this->io->write("Creating temporary fake git repo in $path to apply patch", true, IOInterface::VERBOSE);
+
+        // Create a fake Git repo -- just enough to make Git think it's looking at a real repo.
+        $dirs = [
+            $path . '/.git',
+            $path . '/.git/objects',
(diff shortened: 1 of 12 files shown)


src/Patcher/GitInitPatcher.php
```diff
@@ -0,0 +1,42 @@
...
+    public function apply(Patch $patch, string $path): bool
+    {
+        // If the target is already a git repo, the standard Git patcher can handle it.
+        if (is_dir($path . '/.git')) {
+            return false;
+        }
+
+        $this->io->write("Creating temporary fake git repo in $path to apply patch", true, IOInterface::VERBOSE);
+
+        // Create a fake Git repo -- just enough to make Git think it's looking at a real repo.
```
deviantintegral, line 19: 👋 

I think by not calling `git init` that it relies too much on the internal implementation of the `.git` directory. What if a later version requires some fourth file?

I suppose if this is documented somewhere by git it would be OK.

cweagans: I think you're right. Changed this to just use `git init` in a14cba705c2d61d7ac73bb9ed00576cda11f0f30.

PR #537, opened 2024-01-03
Properly handle removal of all patches

### LINKED ISSUE #525: Removing package for re-install and re-patch does not trigger if patches definition is empty (1.7.3)
### Verification

- [X] I have updated Composer to the most recent stable release (`composer self-update`)
- [X] I have updated Composer Patches to the most recent stable release (`composer update cweagans/composer-patches`)
- [X] I am using one of the supported PHP versions (8.0+)
- [X] I have searched existing issues _and_ discussions for my problem.
- [X] My problem is _not_ addressed in the [troubleshooting guide](https://docs.cweagans.net/composer-patches/troubleshooting/guide).

### What were you trying to do (and why)?

This is using composer-patches 1.7.3

Removing and re-installing patched modules does not behave correctly when no patches are provided in the `composer.json` or external patches JSON file.

For example, assuming the following starting state, the below patches have been applied:

```json
{
    "patches": {
        "acme/module-1": {
            "mock-patch-1": "mock-patch-1.patch"
        },
        "acme/module-2": {
            "mock-patch-2": "mock-patch-2.patch"
        }
    }
}
```

If the user wants to remove both `mock-patch-1` and `mock-patch-2`, and do so by modifying the JSON to the following:
```json
{
    "patches": {}
}
```

Then running `composer install`, the previously applied patches would not be removed and the output would be as follows:
```
Gathering patches from patch file.
No patches supplied.
```

This differs from expected behavior, for example if we were instead removing only `mock-patch-2` the following JSON could be provided:

```json
{
    "patches": {
        "acme/module-1": {
            "mock-patch-1": "mock-patch-1.patch"
        }
    }
}
```

Running `composer install` now would re-install `acme/module-2` and remove `mock-patch-2`, and the output would be as follows:
```
Gathering patches from patch file.
Removing package acme/module-2 so that it can be re-installed and re-patched.
  - Removing acme/module-2 (1.0.0)
```

The expectation is this same process would be applied to an empty `patches` definition too. I believe, but haven't verified that this did used to be the case.

I have worked around this issue temporarily by providing fake empty patch in my patches JSON that targets a non-existent module which appears to have the desired behavior.

### What happened? What did you expect to happen?

Expected `composer install` to re-install previously patched modules, restoring them to the un-patched state.

`composer install` returned "no patches supplied" but did not re-install modules that were previously patched. Previously applied patches remained in place.

dotdash:
## Description

Currently, if there are no patches provided, we assume that there's
nothing to be done, but that's wrong in case that any installed package
has patches applied, as they need to be re-installed without the
patches now. So the early return needs to be removed.

Fixes #525

## Related tasks

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -123,11 +123,6 @@ class Patches implements PluginInterface, EventSubscriberInterface {
         $tmp_patches = $this->arrayMergeRecursiveDistinct($tmp_patches, $patches);
       }
 
-      if ($tmp_patches == FALSE) {
-        $this->io->write('<info>No patches supplied.</info>');
-        return;
-      }
-
       // Remove packages for which the patch set has changed.
       $promises = array();
       foreach ($packages as $package) {


cweagans:

PR #538, opened 2024-01-03
Properly handle all patches being ignored

dotdash:
## Description

The check to determine if there are any patches present for a given
package currently yields `true` for an empty set of patches, which can
happen if all patches provided by a package are marked as ignored. This
causes the affected package to be re-installed everytime, because
`$has_patches` is true while `$has_applied_patches` is false. The logic
for `$has_patches` should also check for the number of patches being
greater than 0, just like it is for `$has_applied_patches`.

## Related tasks


cweagans:

PR #541, opened 2024-01-31
FR - Allow version based patches with 2.x.

rajeshreeputra:
## Description
Capability to apply version specific patches.

Relates to/closes #issueID.

## Related tasks

- [ ] Tests have been added

## Other notes
 example:
```
{
  "patches": {
    "drupal/donotuse": {
      "1.0.8": {
        "Do not use": "https://gist.githubusercontent.com/rajeshreeputra/44c76e98de7831c65a0dcd2558a3129b/raw/b46617ed4c15189e627b69a73d16f383859b9dda/do-not-use-1-0-8.patch"
      },
      "1.0.9": {
        "Do not use": "https://gist.githubusercontent.com/rajeshreeputra/44c76e98de7831c65a0dcd2558a3129b/raw/a041fef3ed7cf22fb7e1dbb366b8d56a72892d07/do-not-use-1-0-9.patch"
      }
    }
  }
}
```

In above example this it applies the first patch when drupal/donotuse is required/updated to 1.0.8 version and second patch applies when drupal/donotuse is required/updated to 1.0.9 version.

**In-progress:**

- [ ] Wildcard version patching like:
   - 1.0.x
   - 1.x
- [ ]  Add test case coverage.

### DIFF
--- a/src/Resolver/ResolverBase.php
+++ b/src/Resolver/ResolverBase.php
@@ -94,18 +94,47 @@ abstract class ResolverBase implements ResolverInterface
                 $temporary_patch_list = [];
 
                 foreach ($patch_defs as $description => $url) {
-                    $patch = new Patch();
-                    $patch->package = $package;
-                    $patch->url = $url;
-                    $patch->description = $description;
-
-                    $temporary_patch_list[] = $patch;
+                    if (is_array($url)) {
+                      foreach ($url as $patchdescription => $patchurl) {
+                        $temporary_patch_list[] = $this->getPatches($package, $patchdescription, $patchurl, $description);
+                      }
+                    }
+                    else {
+                      $temporary_patch_list[] = $this->getPatches($package, $description, $url);
+                    }
                 }
-
                 $patches[$package] = $temporary_patch_list;
             }
         }
 
         return $patches;
+    }
+
+    /**
+     * Helper function to create patch object.
+     *
+     * @param array $package
+     *   The package name.
+     * @param array $description
+     *   The patch description.
+     * @param string $url
+     *   The patch url.
+     * @param string $version
+     *   The version of package.
+     *
+     * @return Patch $patches
+     *   An array of Patch objects.
+     */
+    public function getPatches(string $package, string $description, string $url, string $version = ''): Patch
+    {
+        $patch = new Patch();
+        $patch->package = $package;
+        $patch->url = $url;
+        $patch->description = $description;
+        if (!empty($version)) {
+            $patch->version = $version;
+        }
+
+        return $patch;
     }
 }
--- a/src/Patch.php
+++ b/src/Patch.php
@@ -29,6 +29,13 @@ class Patch implements JsonSerializable
     public string $url;
 
     /**
+     * The version of package.
+     *
+     * @var string $version
+     */
+    public string $version;
+
+    /**
      * The sha256 hash of the patch file.
      *
      * @var ?string sha256
--- a/src/Resolver/PatchesFile.php
+++ b/src/Resolver/PatchesFile.php
@@ -7,6 +7,8 @@
 
 namespace cweagans\Composer\Resolver;
 
+use Composer\Composer\InstalledVersions;
+use Composer\Package\Version\VersionParser;
 use cweagans\Composer\Patch;
 use cweagans\Composer\PatchCollection;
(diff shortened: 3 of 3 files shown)


cweagans:
Before you get too much further with this, you should probably put this in its own plugin in a separate repo. Let me know if I can help with this!

rajeshreeputra:
Yes, please, that would be great idea! thank you! Always appreciate your help!

src/Resolver/ResolverBase.php
```diff
@@ -94,11 +94,15 @@ public function findPatchesInJson(array $patches): array
                 $temporary_patch_list = [];
 
                 foreach ($patch_defs as $description => $url) {
-                    $patch = new Patch();
-                    $patch->package = $package;
-                    $patch->url = $url;
-                    $patch->description = $description;
-
+                    $patch = [];
```
deepakmishra2, line 97:
```suggestion
                    if (is_array($url)) {
                      foreach ($url as $patchdescription => $patchurl) {
                        $patch = $this->getPatches($package, $patchdescription, $patchurl, $description);
                        $temporary_patch_list[] = $patch;
                      }
                    }
                    else {
                      $patch = $this->getPatches($package, $description, $url);
                      $temporary_patch_list[] = $patch;
                    }
```

src/Resolver/ResolverBase.php
```diff
@@ -94,11 +94,15 @@ public function findPatchesInJson(array $patches): array
...
+                    if (is_array($url)) {
+                      foreach ($url as $patchdescription => $patchurl) {
+                        $patch[] = $this->getPatches($package, $patchdescription, $patchurl, $description);
+                      }
+                    }
+                    else {
+                      $patch = $this->getPatches($package, $description, $url);
+                    }
```
deepakmishra2, line 105:
```
if (is_array($url)) {
                      foreach ($url as $patchdescription => $patchurl) {
                        $patch = $this->getPatches($package, $patchdescription, $patchurl, $description);
                        $temporary_patch_list[] = $patch;
                      }
                    }
                    else {
                      $patch = $this->getPatches($package, $description, $url);
                      $temporary_patch_list[] = $patch;
                    }
                    
```
this code should be moved inside foreach and else condition, as $patch is a type of patch object and we are overriding it to convert it into array.

deepakmishra2: also i think we should handle exceptions for wrong urls
`if ($host === NULL) {
          $valid_patches_file = file_exists(realpath($patches_file)) && is_readable(realpath($patches_file));
        }
        else {
          $array = get_headers($patches_file);
          $string = $array[0];

          // 404 for error, 200 for no error
          if(!strpos($string, "200")) {
            return;
          }
        }`

deepakmishra2: also i think we should handle exceptions for wrong urls

```
if ($host === NULL) {
          $valid_patches_file = file_exists(realpath($patches_file)) && is_readable(realpath($patches_file));
        }
        else {
          $array = get_headers($patches_file);
          $string = $array[0];

          if(!strpos($string, "200")) {
            return;
          }
        }
```

cweagans:
This needs to go into a separate plugin entirely. I'm not going to merge this functionality directly into composer-patches directly.

You need to build a _new_ Composer plugin that has a `ResolverProvider`. This is what lets your plugin discover new patches to apply. So you might have a `versioned_patches.json` or something along those lines that has the kind of patch definitions you have above. Your resolver should extend https://github.com/cweagans/composer-patches/blob/main/src/Resolver/ResolverBase.php and in the `resolve` method, you'd call `$collection->addPatch($patch)` for each one you want to add.

In your projects, you'd need to require `cweagans/composer-patches` _and_ e.g. `rajeshreeputra/composer-patches-versioned-patches-resolver` (or whatever you decide to call your project).

See https://docs.cweagans.net/composer-patches/api/overview/ for more information. Feel free to open a discussion in this repo to ask any questions you have about extending composer-patches!

rajeshreeputra:
@cweagans, agree with you, created separate plugin [composer-dynamic-patching](https://github.com/rajeshreeputra/composer-dynamic-patching) by extending [composer-patches](https://github.com/cweagans/composer-patches) [2.x](https://github.com/cweagans/composer-patches/releases/tag/2.0.0-beta1) version.

PR #543, opened 2024-02-08
Prioritize env vars over composer.json setting

chrfritsch:
## Description

Currently, it's impossible to overwrite the composer.json settings `composer-exit-on-patch-failure` and `composer-patches-skip-reporting` with their corresponding env vars.

This PR allows us to easily modify composer-patches behavior, regardless of the settings of the composer.json. This can be very useful in CI pipelines for example.

## Related tasks

- [ ] Tests have been added

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -287,8 +287,8 @@ class Patches implements PluginInterface, EventSubscriberInterface {
 
     // Check if we should exit in failure.
     $extra = $this->composer->getPackage()->getExtra();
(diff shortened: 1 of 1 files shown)


cweagans:

chrfritsch:
Hi, when will 2.0 be stable?

cweagans:
It's more stable than 1.x right now, IMO. I'm going to tag an RC of 2.x in the near-ish future. I'm really just waiting on feedback on the 2.x version at this point.

PR #547, opened 2024-02-13
Dependency patch resolution

cweagans:
## Related tasks

- [x] Tests have been added

### DIFF
--- a/docs/api/patches-lock-json.md
+++ b/docs/api/patches-lock-json.md
@@ -19,4 +19,4 @@ If the [`COMPOSER`]({{< relref "../usage/configuration.md#composer" >}}) environ
 ```
 * Each patch definition will look like the [expanded format]({{< relref "../usage/defining-patches.md#expanded-format" >}}) that users can put into their `composer.json` or external patches file.
 * No _removals_ or _changes_ will be made to the patch definition object. _Additional_ keys may be created, so any JSON parsing you're doing should be tolerant of new keys.
-* The `extra` object in each patch definition may contain a number of attributes set by other projects or by the user and should be treated as free-form input.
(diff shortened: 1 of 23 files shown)


mbrodala:
How is this different from the support removed with https://github.com/cweagans/composer-patches/pull/267 in beta1? Especially that dependency patches are applied by default doesn't seem to have changed. Which would still be a security issue.

So unless I'm missing something: please make applying patches from dependencies an explicit opt in. Always, no matter if there is an own `patches` section. If not enabled a notice for each "missed" patch could raise attention.

garvinhicking:
+1 (I agree with @mbrodala and just learned today of this hidden feature, wasn't aware of it. I really feel this is dangerous as an opt-out and needs to be opt-in, but haven't further investigated this)

mbrodala:
In the meantime I was pointed at the `disable-resolvers` option added here: https://github.com/cweagans/composer-patches/pull/212 ([docs](https://docs.cweagans.net/composer-patches/usage/configuration/#disable-resolvers))

For the reasons mentioned above this option should default to `["\\cweagans\\Composer\\Resolver\\Dependencies"]` instead of `[]`, thus disable dependency patching by default. But honestly, this is no good pattern either:

1. Enabling dependency patching requires you to explicitly use `"disable-resolvers": []` which is completely ambiguous.
2. If you want to disable e.g. the `patches.json` you are likely to use `"disable-resolvers": ["\\cweagans\\Composer\\Resolver\\PatchesFile"]`; you now unintentionally allowed dependency patches

So IMO instead of having a list of resolvers to disable, this should be a list of plain toplevel options instead:

```json
    "extra": {
        "composer-patches": {
            "enable_root_patches": true,
            "enable_root_patches_file": true,
            "enable_dependency_patches": false,
        }
    }
```

(Yes, these would be the sane defaults.)

PR #584, opened 2024-06-24
Made patches.lock.json configurable #533

### LINKED ISSUE #533: [2.*] Make lock file path configurable
### Verification

- [X] I have updated Composer to the most recent stable release (`composer self-update`)
- [X] I have updated Composer Patches to the most recent stable release (`composer update cweagans/composer-patches`)
- [X] I am using one of the supported PHP versions (8.0+)
- [X] I have searched existing issues _and_ discussions for my idea.

### Is your feature request related to a problem?

It's of course not a show stopper but it would be nice if we could also configure the name of the lock file in 2.*. I have a project that also contains Node dependencies and thus the name `patches.json` is ambigious. It would be great if we could change it the same way like we can do it with the patches file.

### Describe your proposed solution(s)

A new property `lock-file` that specifies the name of the lock file. 

### Describe alternatives

_No response_

### Additional context

I could give it a try and raise a PR if you want.

MakerTim:
## Description

Relates to/closes #533

## Related tasks

- [x] Tests have been added

## Other notes

I'm new to the composer-plugin world, so have no clue that this is even the correct path.
Before writing docs and tests can someone confirm this is the correct path?

### DIFF
--- a/src/Plugin/Patches.php
+++ b/src/Plugin/Patches.php
@@ -86,23 +86,6 @@ class Patches implements PluginInterface, EventSubscriberInterface, Capable
     protected JsonFile $lockFile;
 
     /**
-     * Get the path to the current patches lock file.
-     */
-    public static function getPatchesLockFilePath(): string
-    {
-        $composer_file = \Composer\Factory::getComposerFile();
-
-        $dir = dirname(realpath($composer_file));
-        $base = pathinfo($composer_file, \PATHINFO_FILENAME);
-
-        if ($base === 'composer') {
-            return "$dir/patches.lock.json";
-        }
-
-        return "$dir/$base-patches.lock.json";
-    }
-
-    /**
      * Apply plugin modifications to composer
      *
      * @param Composer $composer
@@ -115,12 +98,6 @@ class Patches implements PluginInterface, EventSubscriberInterface, Capable
         $this->executor = new ProcessExecutor($this->io);
         $this->patches = array();
         $this->installedPatches = array();
-        $this->lockFile = new JsonFile(
-            static::getPatchesLockFilePath(),
-            null,
-            $this->io
-        );
-        $this->locker = new Locker($this->lockFile);
         $this->configuration = [
             'disable-resolvers' => [
                 'type' => 'list',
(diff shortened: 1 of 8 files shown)


src/Plugin/Patches.php
```diff
@@ -146,12 +123,44 @@ public function activate(Composer $composer, IOInterface $io): void
...
+    {
+        $composer_file = \Composer\Factory::getComposerFile();
+
+        $dir = dirname(realpath($composer_file));
+        $base = pathinfo($composer_file, \PATHINFO_FILENAME);
+
+        $lock_file = $this->getConfig('patches-lock-file');
+        if (!is_string($lock_file)) {
+            $lock_file = 'patches-lock-file';
+        }
```
cweagans, line 157: This block isn't necessary anymore -- composer configurable plugin handles setting a default for you, so if it's not specified, it's whatever is hardcoded in the config above.

src/Plugin/Patches.php
```diff
@@ -146,12 +123,44 @@ public function activate(Composer $composer, IOInterface $io): void
...
+
+        if ($base === 'composer') {
+            return "$dir/$lock_file";
+        }
+
+        return "$dir/$base-$lock_file";
```
cweagans, line 163: Right now, if you set `COMPOSER=composer-foo.json` and run any of the composer patches commands, you'll get `composer-foo-patches.lock.json`. If we set `patches-lock-file` to `asdf.lock.json` and _also_ set `COMPOSER=composer-foo.json`, I think we'll get `composer-foo-asdf.lock.json`.

This isn't incorrect, but it _is_ something that docs and tests will need to account for. Just flagging for visibility.

src/Plugin/Patches.php
```diff
@@ -85,23 +85,6 @@ class Patches implements PluginInterface, EventSubscriberInterface, Capable
 
     protected JsonFile $lockFile;
 
-    /**
-     * Get the path to the current patches lock file.
-     */
-    public static function getPatchesLockFilePath(): string
```
cweagans, line 91: I can't remember why this was a static function. Might be worth checking if this breaks any of the tests.

cweagans: Ah, yeah. https://github.com/cweagans/composer-patches/actions/runs/9645839515/job/26616352752?pr=584#step:5:137

MakerTim: By not calling the method while registering, its fixed.
Parsing extra's just for the description is not worth it?

cweagans:
Converted this to a draft until docs and tests are added + review feedback is captured. Thanks for your work on this so far!

MakerTim:
Ha @cweagans 

Did write the tests and docs
Anything else?

Issue #551, opened 2024-02-16
Dependency patch resolution: Only allow patches from dependencies you trust


RobinHoutevelts:
### Verification

- [X] I have updated Composer to the most recent stable release (`composer self-update`)
- [X] I have updated Composer Patches to the most recent stable release (`composer update cweagans/composer-patches`)
- [X] I am using one of the supported PHP versions (8.0+)
- [X] I have searched existing issues _and_ discussions for my idea.

### Is your feature request related to a problem?

I only allow patches from sources I trust.
By default `cweagans/composer-patches` seems to apply patches from any of my dependencies.

### Describe your proposed solution(s)

```json
{
   // don't allow patches from anyone else
   "ignore-dependency-patches": ["*"],
   // list of repositories you trust to apply patches from
   "dependency-patches": ["my-org/drupal-patches"]
}
```

### Additional context

You will have less bug reports related to Dependency patch resolution if users are forced to know where they want to apply patches from.

However, I don't mind it -not my project- but I would need a way to opt-out and instead define the sources I trust myself.

cweagans:
I'm on board. Maybe we can add another config value 'allow-dependency-patches' and have that act as an allow list if set. I don't want to have to set both. Having both set should probably be an error.

RobinHoutevelts:
Thanks!

> Having both set should probably be an error.

Yes, that would also make this much better to maintain 👍

RobinHoutevelts:
If I want to disable dependency patch resolution altogether: 
Would setting `allow-dependency-patches: []` (empty list) achieve that?

cweagans:
You can just disable the resolver - no need to try to make an allow list work for that :)

RobinHoutevelts:
@cweagans I'd be happy to take a stab at adding `allow-dependency-patches`.
But I'm currently busy for the next two weeks.

cweagans:
No worries. Same thing here - I'll update here if I beat you to it!

cweagans:
Correct. If the option is NOT defined, all dependencies will be scanned for patches. If it IS defined, only the listed packages will be scanned.

Natshah:
I'm so interested to having this option, something like allow-plugins but it could be called `allowed-patches` or `allowed-patchers` 
as we do have `allowed-packages` for `drupal-scaffold`
```
      "allowed-packages": [
        "drupal/core", // Only listing to demo
        "vardot/varbase"
      ],
```
as
```
      "allow-dependency-patches": [
[...]

PR #588, opened 2024-08-02
[#551] Dependency patch resolution: Only allow patches from trusted d…

metalbote:
…ependencies

## Description

Relates to #551.
Added config key as decribed in the issue, added handling for whitelist, while not breaking current behavior if new config key is omitted.

## Related tasks

- [ ] Tests have been added

### DIFF
--- a/src/Plugin/Patches.php
+++ b/src/Plugin/Patches.php
@@ -146,6 +146,10 @@ class Patches implements PluginInterface, EventSubscriberInterface, Capable
                 'type' => 'string',
                 'default' => 'patches.json',
             ],
+            "allow-dependency-patches" => [
+                'type' => 'list',
+                'default' => null,
+            ],
             "ignore-dependency-patches" => [
                 'type' => 'list',
                 'default' => [],
--- a/src/Resolver/Dependencies.php
+++ b/src/Resolver/Dependencies.php
@@ -24,21 +24,42 @@ class Dependencies extends ResolverBase
             return;
         }
 
-        $this->io->write('  - <info>Resolving patches from dependencies.</info>');
-
+        // Using both config keys, if $allowed_dependencies is set, it takes precedence
+        $allowed_dependencies = $this->plugin->getConfig('allow-dependency-patches');
         $ignored_dependencies = $this->plugin->getConfig('ignore-dependency-patches');
+
+        // First check, if we do allow dependency patches at all.
+        if ($allowed_dependencies === []) {
+            $this->io->write('  - <info>No patches from dependencies are allowed.</info>');
+            return;
+        }
+
+        $this->io->write('  - <info>Resolving patches from dependencies.</info>');
 
         $lockdata = $locker->getLockData();
         foreach ($lockdata['packages'] as $p) {
-            // If we're supposed to skip gathering patches from a dependency, do that.
-            if (in_array($p['name'], $ignored_dependencies)) {
-                continue;
+            $allowed = in_array($p['name'], $allowed_dependencies ?? []);
+            $ignored = in_array($p['name'], $ignored_dependencies);
+
+            // Allowed dependencies is not set in composer.json, and we're not
+            // supposed to skip gathering patches from this dependency
+            if (is_null($allowed_dependencies) && !$ignored) {
+                $this->lookForPatches($p, $collection);
             }
 
-            // Find patches in the composer.json for dependencies.
-            if (!isset($p['extra']) || !isset($p['extra']['patches'])) {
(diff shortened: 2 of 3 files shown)


src/Plugin/Patches.php
```diff
@@ -146,6 +146,10 @@ public function activate(Composer $composer, IOInterface $io): void
                 'type' => 'string',
                 'default' => 'patches.json',
             ],
+            "allow-dependency-patches" => [
+                'type' => 'list',
+                'default' => null,
```
cweagans, line 151: This should probably default to `[]`

src/Resolver/Dependencies.php
```diff
@@ -24,21 +24,42 @@ public function resolve(PatchCollection $collection): void
...
         }
 
-        $this->io->write('  - <info>Resolving patches from dependencies.</info>');
-
+        // Using both config keys, if $allowed_dependencies is set, it takes precedence
+        $allowed_dependencies = $this->plugin->getConfig('allow-dependency-patches');
         $ignored_dependencies = $this->plugin->getConfig('ignore-dependency-patches');
 
+        // First check, if we do allow dependency patches at all.
+        if ($allowed_dependencies === []) {
```
cweagans, line 32: Maybe `empty($allowed_dependencies)` in light of defaulting the value to `[]`?

src/Resolver/Dependencies.php
```diff
@@ -24,21 +24,42 @@ public function resolve(PatchCollection $collection): void
...
+
+        $this->io->write('  - <info>Resolving patches from dependencies.</info>');
+
         $lockdata = $locker->getLockData();
         foreach ($lockdata['packages'] as $p) {
-            // If we're supposed to skip gathering patches from a dependency, do that.
-            if (in_array($p['name'], $ignored_dependencies)) {
-                continue;
+            $allowed = in_array($p['name'], $allowed_dependencies ?? []);
+            $ignored = in_array($p['name'], $ignored_dependencies);
```
cweagans, line 42: Maybe we should simplify this a bit and disallow setting both allowed/ignored dependencies. You can either say "I only want patches from these listed deps" _or_ "I want patches from all deps except these specific ones"

It doesn't make sense to say "I want patches from all deps except these specific ones, but only from these listed deps"

metalbote: We have 4 possibilities, if we remain the ignored dependencies possibility:
a) Current behavior: i have no allow-dependency-patches, and no ignored-dependency-patches, i wanted to preserve that all dependent patches will be used. (Case null)
b) Current behavior: I have no allow-dependency-patches, and some ignored-dependency-patches, all dependent patches should be installed except those listed in ignored. (Case null)
c) New behavior: I do not want any dependency patches, so allow-dependency should be a empty list. (Case [] )
d) New behavior: I want do allow some packages, but because of composer-merge i forgot, i have forbitten in another file package x/y to dependent patches. So this will not break existing installations but is really an enhancement without breaking.

If its okay to break current behavior, we could do differently. But i wanted to keep existing behavior, and wanted to keep expected behavior, like an empty list of allowed dependency implicit none is allowed, and if i ignore something, it will be ignored.

src/Plugin/Patches.php
cweagans: There aren't existing projects that are using 2.x with this functionality, as this functionality does not yet exist.

2.x uses explicit configuration options and very little is implied by other options. If you don't want dependency patch resolution at all, then you simply disable that resolver. This option will be for _changing_ the behavior, not disabling it.

src/Resolver/Dependencies.php
```diff
@@ -24,21 +24,42 @@ public function resolve(PatchCollection $collection): void
...
-        $this->io->write('  - <info>Resolving patches from dependencies.</info>');
-
+        // Using both config keys, if $allowed_dependencies is set, it takes precedence
+        $allowed_dependencies = $this->plugin->getConfig('allow-dependency-patches');
         $ignored_dependencies = $this->plugin->getConfig('ignore-dependency-patches');
 
+        // First check, if we do allow dependency patches at all.
+        if ($allowed_dependencies === []) {
+            $this->io->write('  - <info>No patches from dependencies are allowed.</info>');
+            return;
```
tscni: I wonder if it would be more appropriate to error, if a patch from a dependency is not allowed.
Without applying the specified patches, a dependency might not work as expected.

codebymikey:
As I can't work on the current fork, I've created a PR in #719 that should hopefully be able to get this over the line fairly soon.

PR #612, opened 2025-02-02
Report if 'patch' command does not exist.

donquixote:
## Description

Report if 'patch' command does not exist.

## Related tasks

(this needs to be refined a bit, I think.)

- [ ] Tests have been added

## Other notes

See e.g. https://github.com/ddev/ddev/issues/809
For me this happened when suddenly patches no longer applied, and my first assumption was that my patches are no longer compatible. Instead, the updated ddev web package no longer had 'patch' command available.


cweagans:
Thank you, but no. No longer making changes to 1.x. 2.x relies on Git.

PR #618, opened 2025-03-18
prevent reverse & ignore prev. applied patch

xerc:
https://github.com/cweagans/composer-patches/discussions/567


cweagans:
Thank you, but I'm not accepting changes to 1.x at this point.

PR #621, opened 2025-03-21
[2.x] check & ignore prev. applied patch

xerc:
`git` compatibility-port from https://github.com/cweagans/composer-patches/pull/618/files


xerc:
@cweagans "Thank you, but I'm not accepting changes to 1.x at this point." - this is v2 (ref.@ title)

xerc:
this mod. checks if a PATCH has been applied (maybee with an update or manually)
and therefore does not fail the remaining queue ->  the first `verbose` moved blow

PR #627, opened 2025-05-07
Extended path handling for local patching due to rate limiting issues on github

das-peter:
## Description

This is an attempt to find a sane workaround for issues like this: https://github.com/orgs/community/discussions/157887
My assumption is that we'll probably see more similar issues as long as the LLM craze with all its wild west bots is active.

Unfortunately we're currently not in a position to upgrade to version 2.x where we most likely could workaround all of this using the new API.
An attempt of using the events available in 1.x with a custom stream wrapper to resolve local patches ended with the fact that PHP's [realpath](http://php.net/realpath)() does not support stream wrappers. I'll keep digging to find a way around but by now I'm not sure if this is worth it in comparison with the relatively small changes necessary directly in this package.

The implemented approach of having project root relative paths should allow to keep a stable ignored patches integration too.

We might want to update the documentation even though I think for 1.x it doesn't state anything about the local patch handling - for 2.x there's this: https://docs.cweagans.net/composer-patches/usage/defining-patches/#dependencies

If this change goes into 1.x I'll happily provide a follow-up for 2.x to ensure feature parity.

## Related tasks

- [ ] Tests have been added

### DIFF
--- a/src/Patches.php
+++ b/src/Patches.php
@@ -110,6 +110,7 @@ class Patches implements PluginInterface, EventSubscriberInterface {
       foreach ($packages as $package) {
         $extra = $package->getExtra();
         if (isset($extra['patches'])) {
+          $extra['patches'] = $this->resolveRelativePathPatches($package, $extra['patches']);
           if (isset($patches_ignore[$package->getName()])) {
             foreach ($patches_ignore[$package->getName()] as $package_name => $patches) {
               if (isset($extra['patches'][$package_name])) {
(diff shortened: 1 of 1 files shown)


das-peter:
@jankowa You're still fetching patches from remote. What this patch does is allow packagesto keep the patches in a local folder in a way that works for dependencies as well.
In our setup we have projects which have a dependency which applies patches - without this patch here that dependency could not define local patch files because the paths wouldn't be resolved properly.

mxr576:
@cweagans What do you think - could this be added to the 1.x branch?

Even setting aside the recent GitHub rate-limiting changes that may have triggered this need, the broader concept of allowing package releases to bundle necessary patches for their dependencies seems like a solid security enhancement. It also offers a practical alternative to forking and manually mirroring every PR/CR/MR diff as a static link.

More context from the Drupal documentation:
https://www.drupal.org/docs/develop/git/using-gitlab-to-contribute-to-drupal/downloading-a-patch-file#s-use-downloaded-patch-files

das-peter:
For anyone stumbling over this - we've created a helper package to download remote patches: https://github.com/CandoImage/composer-patches-inliner

cweagans:
No, sorry. 1.x isn't really changing at this point. 2.x will likely solve your problem, as a given patch should only ever be downloaded once (provided you keep the cache directory).

PR #631, opened 2025-05-14
Skip tests on "429 Too Many Requests" response

zonuexe:
## Description

For the reasons mentioned in https://github.com/cweagans/composer-patches/pull/627, the tests for composer-patches themselves will fail, which looks bad, so we will detect the rate limit and skip the tests.

> 

## Related tasks

- [ ] Tests have been added


cweagans:
Thank you, but no. Since the tests in main are actually useful, I'd prefer that they either run or they complain loudly so that they can be fixed.

PR #637, opened 2025-08-11
Fix freeform patcher applying patch during dry run

darrenoh:
## Description

The freeform patcher was using the regular args for the dry run instead of the dry run args. If dry run args were provided, the patch was applied twice.


cweagans:
Thank you!

Issue #561, opened 2024-03-20
Is patch provenance reported correctly?


cweagans:
### Verification

- [X] I have updated Composer to the most recent stable release (`composer self-update`)
- [X] I have updated Composer Patches to the most recent stable release (`composer update cweagans/composer-patches`)
- [X] I am using one of the supported PHP versions (8.0+)
- [X] I have searched existing issues _and_ discussions for my problem.
- [X] My problem is _not_ addressed in the [troubleshooting guide](https://docs.cweagans.net/composer-patches/troubleshooting/guide).

### What were you trying to do (and why)?

Patches lock supplied in https://github.com/cweagans/composer-patches/issues/559 doesn't seem to be reporting provenance correctly. Patches applied to drupal/ctools are reported as being _from_ drupal/ctools, but the upstream repo is not doing that

### What happened? What did you expect to happen?

a

### Contents of `composer.json`

```shell
a
```

### Contents of `patches.lock.json`

```shell
a
```

### Full output of `composer patches-doctor`

```shell
a
```

### Full output of relevant Composer command with the `-vvv` flag added

```shell
a
```

PR #657, opened 2025-11-13
Fix: Wrong provenance for dependency patches

blankse:
## Description

The provenance should be the dependency in which the patches are defined, not the dependency that was patched.

## Related tasks

- [x] Tests have been added

Fixes https://github.com/cweagans/composer-patches/issues/561


blankse:
@cweagans I added a second seeInThisFile method call to test the correct dependency

cweagans:
Thank you!

PR #667, opened 2025-12-16
Multiple patch listing files

NickDickinsonWilde:
## Description
No issue... maybe I should've written an issue before a PR? (but .github/CONTRIBUTING.md is empty so unsure).

Upgrading a pile of Drupal sites to PHP 8.5. Almost all of them need the same 4-8 patches. Most of them still have some PHP 8.4 related patches. Having to merge JSON manually for each site is more work than I think needed. Instead, I want to have multiple patches files...

patches.json (or whatever name) with site specific patches
php84.patches.json with PHP 8.4 related patches for components that most of my sites have
php85.patches.json with PHP 8.5 related patches for common components.

Because packages in the patch file that don't exist on the project are silently ignored that is really convenient and reduces work on similar projects.

Example output:
image

## Related tasks
_Everything here needs to be done if the idea is approved - happy to do so, but wanted to get a POC out for my own personal use/testing_
- [ ] Tests have been added
- [ ] Does not break backwards compatibility _OR_ a BC break has been discussed in the related issue(s). -- in current patch I am messaging that current `patches-file` config key is deprecated. But that should probably be discussed.

### DIFF
--- a/src/Plugin/Patches.php
+++ b/src/Plugin/Patches.php
@@ -146,6 +146,10 @@ class Patches implements PluginInterface, EventSubscriberInterface, Capable
                 'type' => 'string',
                 'default' => 'patches.json',
             ],
+            'patch-files' => [
+                'type' => 'list',
+                'default' => [],
+            ],
             "ignore-dependency-patches" => [
                 'type' => 'list',
                 'default' => [],
--- a/src/Resolver/PatchesFile.php
+++ b/src/Resolver/PatchesFile.php
@@ -18,22 +18,26 @@ class PatchesFile extends ResolverBase
      */
     public function resolve(PatchCollection $collection): void
     {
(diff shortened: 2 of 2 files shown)


cweagans:
I think this should probably be a separate plugin that provides a [resolver](https://docs.cweagans.net/composer-patches/getting-started/terminology/#resolver). The challenging thing here will be figuring out what should end up in the lock file. It's meant to be a comprehensive view of _all_ patches that should be applied to a project. Adding conditional resolution like this might complicate things. Maybe everything goes into the lock file and you add an event handler that suppresses patches that are not relevant to the current project?

cweagans:
> .github/CONTRIBUTING.md is empty

Thanks for flagging this! Fixed in https://github.com/cweagans/composer-patches/commit/6111ede4e7a11518427f372271f7a391a1d80da6

Issue #634, opened 2025-07-10
Allow patches relative to `composer.json`


lkmorlan:
### Verification

- [x] I have updated Composer to the most recent stable release (`composer self-update`)
- [x] I have updated Composer Patches to the most recent stable release (`composer update cweagans/composer-patches`)
- [x] I am using one of the supported PHP versions (8.0+)
- [x] I have searched existing issues _and_ discussions for my idea.

### Is your feature request related to a problem?

Scenario: Project A requires Project B. Project B requires Project C and provides a patch file for it. The path to this patch file, configured in the Project B `composer.json`, will be interpreted relative to Project A rather than to its own `composer.json`. I can get it to work as long as Project B knows at what path Project A will install it. But Project B may not know this.

### Describe your proposed solution(s)

A project should be able to declare that patch file paths are relative to that project's `composer.json`. This could be done by having a token which represents the full path to the directory containing the project's `composer.json`. So, the patches section would look something like this:

```
    "extra": {
        "patches": {
            "example/project": {
                "Comment": "[this-project]/patches/changes.patch"
            }
        }
    }
```

PR #668, opened 2025-12-17
#634 Allow patches relative to composer.json

chx:
## Description

This implements #634 . The issue suggested

```
    "extra": {
        "patches": {
            "example/project": {
                "Comment": "[this-project]/patches/changes.patch"
            }
        }
    }
```

but I felt that's unnecessary, what else could a local path mean but a project relative path? So in the project foo `composer.json` you could find

```
    "extra": {
        "patches": {
            "bar/library": {
                "Comment": "patches/patch-for-bar-library.patch"
            }
        }
    }
```

## Implementation note

We need to download and install the package into a temporary directory because we are pre install so the package is not installed yet (duh) but the patch is needed for hashing.

## Security considerations

I restricted the path to be strictly inside the project although I do not readily see how is that necessary as any random file on the filesystem you'd want to steal is extremely unlikely to be a patch file. But the cost for this is so low I would rather do this.

## Remaining issues

The test suite fails. Running the tests independently they pass. What gives??

## Possible followups

Shouldn't the directly instantiated Downloader class be downloadmanager too?


src/Downloader/ComposerDownloader.php
```diff
@@ -29,7 +32,34 @@ public function download(Patch $patch): void
...
-        $downloader->copy($patch->url, $filename);
+        // Patch inside the package.
+        if (isset($patch->extra[Dependencies::RELATIVE_PATCH_PROVIDER])) {
+            $provider = $patch->extra[Dependencies::RELATIVE_PATCH_PROVIDER];
+            if ($package = $this->getPackage($provider)) {
+                $dir = sys_get_temp_dir() . \DIRECTORY_SEPARATOR . uniqid($provider);
+                try {
+                    $dm = $this->composer->getDownloadManager();
+                    $dm->download($package, $dir);
+                    $dm->install($package, $dir);
```
adam-vessey, line 43: Maybe why this gets dicey between environments (the test suite vs not), but: These are promises, so might have to make sure that they've resolved before making use of their output. Looks like [Composer provides a `SyncHelper`](https://github.com/composer/composer/blob/main/src/Composer/Util/SyncHelper.php) that could be used for (infrequent?) downloads? Or, could possibly naively do a `$this->composer->getLoop()->wait([$promise]]);` kind of thing?


```suggestion
                    SyncHelper::downloadAndInstallPackageSync($this->composer->getLoop(), $this->composer->getDownloadManager(), $dir, $package);
```

might do the trick (assuming an earlier `use Composer\Util\SyncHelper;`)?

src/Downloader/ComposerDownloader.php
```diff
@@ -29,7 +32,34 @@ public function download(Patch $patch): void
...
+                    $newUrl = realpath($dir . \DIRECTORY_SEPARATOR . $patch->url);
+                    if (file_exists($newUrl) && str_starts_with($newUrl, realpath($dir))) {
+                        copy($newUrl, $filename);
+                    }
```
adam-vessey, line 47: Should possibly throw an exception if we don't find the patch?

src/Downloader/ComposerDownloader.php
```diff
@@ -29,7 +32,34 @@ public function download(Patch $patch): void
...
+                } catch (\Exception $e) {
+                    // @TODO anything to do here?
+                } finally {
```
adam-vessey, line 50: Not sure we want to catch any exception here, or just allow it to bubble out to composer to stop the current operation?

src/Downloader/ComposerDownloader.php
```diff
@@ -40,4 +70,19 @@ public function download(Patch $patch): void
...
+        foreach ($this->composer->getRepositoryManager()->getRepositories() as $repository) {
+            try {
+                if ($packages = $repository->findPackages($packageName)) {
+                    return reset($packages);
+                }
+            } catch (\Exception $e) {
+                continue;
+            }
+        }
+        return null;
```
adam-vessey, line 86: Should we more strongly assert that we want this package? Throwing an exception here instead of returning `null`?

src/Downloader/ComposerDownloader.php
```diff
@@ -29,7 +32,34 @@ public function download(Patch $patch): void
             mkdir($patches_dir);
         }
 
-        $downloader->copy($patch->url, $filename);
+        // Patch inside the package.
+        if (isset($patch->extra[Dependencies::RELATIVE_PATCH_PROVIDER])) {
+            $provider = $patch->extra[Dependencies::RELATIVE_PATCH_PROVIDER];
+            if ($package = $this->getPackage($provider)) {
```
adam-vessey, line 38: If `$this->getPackage([...])` just threw an exception (and never returned `null`), we could avoid the `if` here.

PR #713, opened 2026-08-06
Include patch description in no available patcher error message

### LINKED ISSUE #666: Failure no longer shows description
### Verification

- [x] I have updated Composer to the most recent stable release (`composer self-update`)
- [x] I have updated Composer Patches to the most recent stable release (`composer update cweagans/composer-patches`)
- [x] I am using one of the supported PHP versions (8.0+)
- [x] I have searched existing issues _and_ discussions for my idea.

### Is your feature request related to a problem?

In the 2.0 release the description isn't in the output making it more difficult to debug patch failures.

For example, previously patch failures would look something like:
```
  - Applying patches for drupal/core
    https://git.drupalcode.org/project/drupal/-/commit/dne.patch (really cool patch from next release)
   Could not apply patch! Skipping. The error was: Cannot apply patch https://git.drupalcode.org/project/drupal/-/commit/dne.patch
```

Its quite clear this was the "really cool patch"

But now they look like:
```
  - Patching drupal/core
      - Found cached patch at /mnt/ddev-global-cache/composer/patches/c0388b619228608ba69d8303e07789fb3762d41d58ba8aa26d53529ad498022a.patch

In Patches.php line 288:
                                                                                                                            
  No available patcher was able to apply patch https://git.drupalcode.org/project/drupal/-/commit/dne.patch to drupal/core  
                                                                                                                            
```

I often use this to reference the merge/issue or other relevant context so when you see the failure its obvious what failed, what it was used for, where to go to engage with fixing it, etc.

Now, unless its a local patch with a very descriptive file name, its much more difficult to diagnose requiring digging through the patches list.

### Describe your proposed solution(s)

Add the description back some where in the patching or failure output so its clear what is being applied and failed.

### Describe alternatives

Don't know what an alternative would be. This would just be restoring some useful context to developers.

### Additional context

_No response_

wolcen:
## Description

Relates to/closes #666. 😈

## Related tasks

- [X] Documentation has been updated if applicable
- [X] Tests have been added
- [X] Does not break backwards compatibility _OR_ a BC break has been discussed in the related issue(s).

## Other notes

`composer test` tests passing for me, and this is tested by `NoPatchersAvailableCept` and `PatchDoesNotApplyCept`.

Don't think this would break BC (at least I would not expect such output to be guaranteed).

It would appear you can technically not enter a patch description, but I did not do anything special for this (other code doesn't appear to either).

I would lastly note that using `-v` does also tell you the patch name (per same file, line 345), but there's a lot of other noise in that output.


wolcen:
Removed the merge commit...not a big user of the GH UI, didn't know it'd do that instead of a rebase. Anyway...thanks for considering.

